PR #24, opened 2016-01-31
Reimplement  since it's not supported in all libcs

simias:
In particular it seems Mingw doesn't have it even though it's supposed to be POSIX.


Monroe88:
I just added `CXXFLAGS += -D ffs=__builtin_ffs` to the Windows build target and it worked for me without this commit

PR #59, opened 2016-05-30
Move crop overscan to a core option

hizzlekizzle:
and add a couple more options for finer control.

### DIFF
--- a/libretro.cpp
+++ b/libretro.cpp
@@ -2747,6 +2747,8 @@ static void check_variables(bool startup)
    else
      display_internal_framerate = false;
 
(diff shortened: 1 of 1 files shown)


Monroe88:
Might want to remove these two lines to avoid having RetroArch's "crop overscan" option overriding the core option

https://github.com/libretro/beetle-psx-libretro/blob/master/libretro.cpp#L3180

hizzlekizzle:
You're right, I'll zap them real quick. I don't think they interfere, since I changed the variable that triggers the behavior, but they don't need to be there anymore, regardless.

PR #73, opened 2016-07-18
All exceptions are now caught within retro_load_game

sronsse:
Exceptions are now caught while attempting to load content. If one occurs, then an error log message is printed out and control comes back to the frontend.


rz5:
o/

Issue #95, opened 2016-09-28
HW Core Regressions: After committing PGXP changes


saftle:
After PGXP compatibility was merged, certain games show the wrong colors or continuously flash if the Internal Color Depth: 32bpp option is enabled. A good example game is Brave Fencer Musashi & Wild 9.

In this case, the problem is reproducible even when all PGXP relevant core options are disabled. The workaround is to turn off Internal Color Depth: 32bpp and re-enable it. However, the problem will occur every time new textures are loaded, so it's not a permanent fix.

I have reproduced this with the Windows Buildbot build from today.

simias:
I can't seem to find any issue in Brave Fencer Musashi (but I don't know the game very well so I might be missing things). Is there some way to reproduce the issue easily?

saftle:
If you start Brave Fencer Musashi with "Internal Color Depth: 32bpp" it should automatically show texture issues. This is with all core options disabled, shaders disabled, and only only Internal Color Depth: 32bpp being set. The first thing you'll notice is the opening video has weird bars going down the middle, and then more texture issues. If it's not reproducible on your end, I can make a video if need be. Thanks.

saftle:
Found another issue, this time with Croc. Even with every core option disabled (every native option left alone, and all PGXP options disabled), the game continuously flashes in the opening video.

saftle:
This was fixed awhile ago, but accidentally left it open.

Issue #85, opened 2016-08-27
PGXP Support


Papermanzero:
Adding PGXP would increase the image quality
Details: http://ngemu.com/threads/pcsxr-pgxp.186369/

simias:
I haven't looked at PGXP's implementation yet but in order to do that you'd have to add a side-channel in mednafen's GTE, RAM and DMA code to tunnel additional data from the GTE all the way to the GPU.

Here's how I thought it could be implemented, in theory, in case somebody else wants to give it a try:

## GTE

Here you can see where the perspective division is done in the GTE: https://github.com/libretro/beetle-psx-libretro/blob/master/mednafen/psx/gte.cpp#L1065-L1070

All the 16bit of fractional precision are lost in the bitshift before the result is stored in the GTE's `XY_FIFO` register, effectively truncating to the previous pixel. This is what causes a lot of jitter in PSX graphics, you don't have any subpixel precision.

At the same time you can see that the code doesn't store the `z` value anywhere, it's used for the perspective division and then discarded. Without this value in the GPU we can't do perspective-correct texture mapping which results in warping when viewing polygons not directly facing the camera.

In order to solve these issues we have to keep the enhanced precision x and y as well as the z value somewhere. For instance we could change `XY_FIFO` to hold those new "increased precision" variables alongside the regular integer XY values.

## RAM

Once that's working you need to patch the code that sends GTE data to RAM. This should *mostly* happen [here](https://github.com/libretro/beetle-psx-libretro/blob/master/mednafen/psx/cpu.cpp#L1259). It's possible that some games don't use the SWC2 opcode and instead use MFC2 to move the GTE register value to a main CPU register and then use a regular SW to put it into main RAM or even straight to the GPU but since it's less efficient it's *probably* not super common.

Anyway, here you'd have to special case reads from the XY_FIFO to store the increased precision data in some buffer tied to the main RAM.

## DMA

Then you need to patch the DMA to send the increased precision data from RAM to GPU. The DMA reads the data from RAM [here](https://github.com/libretro/beetle-psx-libretro/blob/master/mednafen/psx/dma.cpp#L432) and sends it to the GPU [here](https://github.com/libretro/beetle-psx-libretro/blob/master/mednafen/psx/dma.cpp#L212). This code will have to be patched to give the GPU a way to access the increased precision data.

## GPU

[...]

PR #98, opened 2016-10-01
Stop quads with partially bad W components from getting rendered.

iCatButler:
Should fix stretched textures in games like Spyro.

### DIFF
--- a/mednafen/psx/gpu.h
+++ b/mednafen/psx/gpu.h
@@ -258,6 +258,7 @@ class PS_GPU
 
       tri_vertex InQuad_F3Vertices[3];
       uint32 InQuad_clut;
+	  bool InQuad_invalidW;
 
       line_point InPLine_PrevPoint;
 
--- a/mednafen/psx/gpu_polygon.cpp
+++ b/mednafen/psx/gpu_polygon.cpp
@@ -443,6 +443,7 @@ INLINE void PS_GPU::Command_DrawPolygon(const uint32_t *cb)
       {
          memcpy(&vertices[0], &InQuad_F3Vertices[1], 2 * sizeof(tri_vertex));
          clut = InQuad_clut;
+		 invalidW = InQuad_invalidW;
          sv = 2;
       }
    }
@@ -510,7 +511,7 @@ INLINE void PS_GPU::Command_DrawPolygon(const uint32_t *cb)
 
    // iCB: If any vertices lack w components then set all to 1
    if (invalidW)
-	   for (unsigned v = sv; v < 3; v++)
+	   for (unsigned v = 0; v < 3; v++)
 		   vertices[v].precise[2] = 1.f;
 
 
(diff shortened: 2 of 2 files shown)


simias:
Why only quads? Couldn't triangle theoretically have the same issue?

twinaphex:
Big breakthrough, the skybox glitches have all been fixed.

twinaphex:
The only issues that remain now with PGXP integration is this -

https://github.com/libretro/beetle-psx-libretro/issues/95

And the other issue is that Ridge Racer Revolution/Rage Racer don't work with PGXP enabled.

twinaphex:
Are some textures now not being properly perspective corrected as a result of this or were there no changes there?

iCatButler:
Triangles were already being caught if a single vertex was missing a valid `W` component and all were set to 1, effectively disabling perspective correction for those triangles.

With quads the final vertex is processed in a second iteration of the function and was not invalidating the other vertices, just itself, hence the triangle it formed part of would stretch.

@twinaphex I think I fixed Ridge Racer with an earlier change. Let me know if it isn't.

simias:
Oh right, I forgot about that.

And I confirm that RRR still freezes as soon as I turn on PGXP during a race even with the very latest commit (b02e5afa8f089d6ed0d54b0ea3b7e4eb44dd8073).

Ridge Racer Type 4 works well however.

twinaphex:
Yeah, Type 4 has always worked with PGXP, it's Ridge Racer Revolution/Rage Racer which are the problem.

iCatButler:
Ah, I think it was the original Ridge Racer I tested, it was crashing in a similar way and now seems to be fixed. I'll have a look at the others.

simias:
@iCatButler I haven't retested it with the latest commit but a few days ago the crash would go away if I removed the call to the PGXP_NCLIP (and used the native result instead), see https://github.com/libretro/beetle-psx-libretro/issues/85#issuecomment-249579767

Hope that helps.

iCatButler:
It looks like I did fix it, but it's a bit temperamental.

The problem isn't actually NCLIP but the values of the vertices being pushed in by the GTE's Transform function. Some of them have a depth value of 0, by which the near plane distance is divided, yielding `+-inf`. This is then multiplied by the X and Y coordinates, corrupting those and leading to the massively wrong NCLIP result further down the line.

Locally I fixed it by using `min()` and `max()` macros to clamp these values to the valid range of -1024 to 1023. On merging the change the `min()` was changed to `std::min<float>`, if I make this change locally then the game freezes due to NCLIP results again.

I feel there's probably some finer point of the standard library I'm forgetting, but I'd expect them to behave pretty much the same.

twinaphex:
You can just use the macros if you prefer, I think we have both MIN and MAX macros too inside the retro_miscellaneous.h header which you could use.

iCatButler:
I'll try dealing with the divide by 0 first, I have a suspicion that the results are undefined, so it's probably safer to do something to avoid it in case this comes back to bite us :)

rz5:
@iCatButler: About that min() to std::min()<float> change - I compile the MSYS2 toolchain on Windows and so does the buildbot. It was failing to compile, so I used the latter function as it was used elsewhere in the same file.

Please don't hesitate in reverting that change though.

iCatButler:
@rz5 no problem, it's a perfectly reasonable change.

If anything I think it's highlighted something the code should be dealing with better, I'm just not sure what the correct behaviour is.

I tried mimicking what the native GTE divide function does, clamping the end results to `0x1FFFF` and setting it explicitly to that on depth values less than near/2, but that causes incorrect culling of near polygons in some cases. So I've settled on clamping depth values to a minimum of near/2 before performing the divide, it's similar to the PCSX-R integration and seems to work well.

PR #101, opened 2016-10-01
Clamp depth values to near/2.f to prevent the projection factor becom…

iCatButler:
…ing too large (or div_by_0)

Should fix RRR.


twinaphex:
Awesome, fixes Ridge Racer Revolution and Rage Racer on my end.

PR #108, opened 2016-10-08
Report that serialized states are variable sized

GregorR:
This updates libretro.h and uses the new serialization quirks API to report variable-sized savestates. This also fixes support for frontends which don't support variable-sized savestates, which at present includes RetroArch. Fixes netplay, probably fixes rewind.


simias:
That's great! I'm going to use this API in my other cores now. Thanks!

PR #116, opened 2016-11-23
(WiiU) added thread wrapper to make it compilable

FIX94:
this will let it at least boot up games though they will run very slowly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -211,13 +211,16 @@ else ifneq (,$(filter $(platform),ngc wii))
 
 # Nintendo WiiU
 else ifeq ($(platform), wiiu)
-	TARGET := $(TARGET_NAME)_libretro_$(platform).a
-	ENDIANNESS_DEFINES += -DHW_RVL -DWIIU
+   TARGET := $(TARGET_NAME)_libretro_$(platform).a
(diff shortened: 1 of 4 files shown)


Alcaro:
> #define static_assert(a,b)

i'll assume that's to work around some quirk in the wiiu headers

(1) comment (2) #undef

> \-ifneq ($(STATIC_LINKING), 1)
> \+#ifneq ($(STATIC_LINKING), 1)

that affects non-wiiu platforms, and probably has unwanted side effects even on that

> static INLINE int pthread_mutex_destroy(pthread_mutex_t *mutex)
> {
>    return 0;
> }

that's a memory leak

> pthread_cond_destroy

that too

> merged

did you even notice those flaws

also not sure if you did this already, but PRs to libretro-common should also be sent to retroarch; we often copy libretro-common around and could end up reverting this unless the master copy is updated as well

aliaspider:
about the ```+#ifneq ($(STATIC_LINKING), 1)``` part, I asked him to comment it out since it was broken in the first place, as it also disabled compiling libkirk, which isn't part of libretro-common.
my guess is that there simply no working statically linked targets for beetle-psx yet.

PR #123, opened 2016-12-03
Some debian patches

sergiobenrocha2:


Makefile
```diff
@@ -390,10 +390,10 @@ else
 endif
 
 %.o: %.cpp
-	$(CXX) -c -o $@ $< $(CXXFLAGS)
+	$(CXX) -c -o $@ $< $(CPPFLAGS) $(CXXFLAGS)
 
 %.o: %.c
-	$(CC) -c -o $@ $< $(CFLAGS)
+	$(CC) -c -o $@ $< $(CPPFLAGS) $(CFLAGS)
```
simias: Are you sure we're supposed to add CPPFLAGS when building C files?

sergiobenrocha2: Yup, it's linkage stuff

http://stackoverflow.com/questions/2754966/cflags-vs-cppflags

simias: Aaah, so it's CPP for C PreProcessor, not C++. Makes sense. Thank you for the clarification.

Issue #136, opened 2016-12-11
[OpenGL] Internal resolution - partial screen


alondero:
When changing the internal resolution settings in the most recent build the portion of the screen that doesn't fall into the bottom left most (assuming) 320x240 is cropped away. This is happening on the <s>Vulkan</s> and OpenGL backends.
![colin mcrae rally 2 0 usa -161211-111017](https://cloud.githubusercontent.com/assets/1269060/21079725/c1bc7dba-bf92-11e6-931d-a706d01722af.png)
My core settings are:
`beetle_psx_hw_renderer = "opengl"
beetle_psx_hw_renderer_software_fb = "enabled"
beetle_psx_hw_adaptive_smoothing = "enabled"
beetle_psx_hw_internal_resolution = "8x"
beetle_psx_hw_filter = "bilinear"
beetle_psx_hw_internal_color_depth = "32bpp"
beetle_psx_hw_scale_dither = "enabled"
beetle_psx_hw_wireframe = "disabled"
beetle_psx_hw_display_vram = "disabled"
beetle_psx_hw_pgxp_mode = "memory + CPU"
beetle_psx_hw_pgxp_caching = "enabled"
beetle_psx_hw_pgxp_texture = "enabled"
beetle_psx_hw_initial_scanline = "0"
beetle_psx_hw_last_scanline = "239"
beetle_psx_hw_initial_scanline_pal = "0"
beetle_psx_hw_last_scanline_pal = "287"
beetle_psx_hw_frame_duping_enable = "enabled"
beetle_psx_hw_widescreen_hack = "enabled"
beetle_psx_hw_crop_overscan = "enabled"
beetle_psx_hw_image_crop = "disabled"
beetle_psx_hw_image_offset = "disabled"
beetle_psx_hw_display_internal_framerate = "disabled"
beetle_psx_hw_analog_calibration = "disabled"
beetle_psx_hw_analog_toggle = "disabled"
beetle_psx_hw_enable_multitap_port1 = "disabled"
beetle_psx_hw_enable_multitap_port2 = "disabled"
beetle_psx_hw_cpu_overclock = "enabled"
beetle_psx_hw_cdimagecache = "enabled"
beetle_psx_hw_skipbios = "enabled"
beetle_psx_hw_use_mednafen_memcard0_method = "libretro"
beetle_psx_hw_enable_memcard1 = "enabled"
beetle_psx_hw_shared_memory_cards = "disabled"
[...]

PR #169, opened 2017-03-26
Convert remaining option names to option_* macros.

rkapl:
The rename was done in 2b2070b, but one file was missed.

Also for some reason fixes the partial screen issue
(https://github.com/libretro/beetle-psx-libretro/issues/136).

I can not tell exactly what caused the issue but it seemed that libretro was getting wrong information about the viewport (the clipping seemed to be on the libretro level).


fr500:
Nice catch... tried to unify it all but it was a mess so I guess I missed some, gonna try now

PR #206, opened 2017-07-26
WinRT and other fixes

albertofustinoni:
- Updated thread library to compile in WinRT
- Fixed bug that left core unusable after unloading a game

### DIFF
--- a/libretro-common/rthreads/rthreads.c
+++ b/libretro-common/rthreads/rthreads.c
@@ -437,11 +437,17 @@ void scond_free(scond_t *cond)
 #ifdef USE_WIN32_THREADS
 static bool _scond_wait_win32(scond_t *cond, slock_t *lock, DWORD dwMilliseconds)
 {
-   static LARGE_INTEGER performanceCounterFrequency = { .QuadPart = 0 };
-
    struct QueueEntry myentry;
    struct QueueEntry **ptr;
+
+#if _WIN32_WINNT >= 0x0500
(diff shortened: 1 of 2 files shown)


twinaphex:
I backported the second commit of yours at least. As for the former one, I need to make sure first that backwards compatibility does not suffer as a result. If so, this new codepath you made might have to be put under ifdefs for RT or something.

albertofustinoni:
The functions I am using are supported from Windows 2000 onwards.

I'd say it is better not to have a second codepath if possible, easier to spot and fix bugs.

For reference look [here](https://msdn.microsoft.com/en-us/library/windows/desktop/ms644904(v=vs.85).aspx) and [here](https://msdn.microsoft.com/en-us/library/windows/desktop/ms644905(v=vs.85).aspx), bottom of the page.

You can see an overview of what I'm doing on [this article](https://msdn.microsoft.com/en-us/library/windows/desktop/dn553408(v=vs.85).aspx)

twinaphex:
Well, there is still the Windows 9x port by @bparker06. Maybe I should invite him into the conversation.

bparker06:
Yes I would suggest wrapping the QPC stuff around `#if _WIN32_WINNT >= 0x0500` since this will break compilation targeting Win9x/NT.

albertofustinoni:
Added preprocessor defines as suggested by @bparker06 and restored the old code for compilation under Win9X.

PR #218, opened 2017-08-14
Fix Playstation executable loading.

alex34567:
Mednafen has functionality to load a playstation executable ([like this](http://psx.amidog.se/doku.php?id=psx:download:cpu#CPU_Test))  instead of an cd.
In Beetle this functionality is implemented but a bug caused the executable to not load.
After a simple patch the functionality now works.


twinaphex:
Thanks a lot!

PR #228, opened 2017-09-03
Add CHD file support

Zapeth:
This is mostly based on the existing implementations in other libretro cores (actually a little trimmed down to only support Mode2/2352 and Audio track images, but other types can still be easily added) and has only been roughly tested for functionality.

However even if currently there might be flaws in the implementation, it is still highly recommended to compress PSX games with this format for any users who are currently using other file formats like PBP (that is proprietary, closed-source and has mostly sloppy to flat out broken converters).


twinaphex:
Hi, there seems to be a merge conflict in Makefile.common that has to be resolved first.

Zapeth:
Yeah, I just wasn't sure if that had to be done on your end or mine (never had to resolve conflicts before), hopefully it should be ok now.

PR #237, opened 2017-10-01
Precache compressed (rather than uncompressed) CHD data

bkoropoff:
This makes the cache disc option more tolerable than my previous attempt.  Loading multi-disc playlists with a cold OS page cache still produces mystifying delays on startup, but at least it doesn't peg a CPU to 100% for over a minute.

I'm still dissatisfied with pre-caching discs to avoid stutter.  You shouldn't need an SSD to be able to keep up with a simulated 2x CD-ROM drive from 1994.  The CD emulation code probably needs to be reworked.


twinaphex:
Yes, the CDROM code in mednafen is inefficient.

PR #242, opened 2017-10-15
Avoid stuttering on CD-ROM sector reads

bkoropoff:
This avoids needing to turn on CD pre-caching to prevent emulation stuttering.  The two changes are fairly simple:

- Instead of synchronously waiting for the background I/O thread and blocking emulation, make the emulated program wait longer and check back later.  This seems unlikely to cause compatibility problems given that real games had to contend with imperfect physical CD-ROM drives -- some of which were being literally melted by the power supply -- but the pre-cache option is available to bypass this if needed.
- Annoyingly, there is `PreSeekHack` method which synchronously reads subchannel data *prior* to emulating the seek time.  Since the CD interface actually reads the entire sector to extract the subchannel data, this defeats many of the benefits of background I/O (speculative read-ahead still helps).  It turns out that the subchannel data is usually computed from scratch anyway, so I refactored it to avoid reading the sector.  It seems reading the subchannel data was intended to provide a hint to the background thread to read the full sector -- this now actually works.

### DIFF
--- a/mednafen/cdrom/CDAccess.cpp
+++ b/mednafen/cdrom/CDAccess.cpp
@@ -59,3 +59,14 @@ CDAccess *cdaccess_open_image(bool *success, const char *path, bool image_memcac
 #endif
    return new CDAccess_Image(success, path, image_memcache);
 }
+
+bool CDAccess::Read_Raw_PW(uint8_t *buf, int32_t lba)
+{
+   uint8 tmpbuf[2352 + 96];
+
+   if (!Read_Raw_Sector(tmpbuf, lba))
+      return false;
+   memcpy(buf, tmpbuf + 2352, 96);
+
+   return true;
+}
--- a/mednafen/cdrom/CDAccess.h
+++ b/mednafen/cdrom/CDAccess.h
@@ -18,6 +18,8 @@ class CDAccess
 
  virtual bool Read_Raw_Sector(uint8_t *buf, int32_t lba) = 0;
 
+ virtual bool Read_Raw_PW(uint8_t *buf, int32_t lba);
+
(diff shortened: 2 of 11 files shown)


twinaphex:
Cool. Are you planning to backport these changes to the other Beetle/Mednafen cores too?

simias:
>Instead of synchronously waiting for the background I/O thread and blocking emulation, make the emulated program wait longer and check back later. [...] but the pre-cache option is available to bypass this if needed.

Just to be clear, is this opt-in or is it enabled by default?

I really think it should be opt-in because it makes the behavior non deterministic: the host capabilities "leak" into the emulated state. If somebody has an issue with this option enabled it might be painful to reproduce. Besides I think speedhacks in general should be opt-in, people should have the more accurate and compatible experience by default and enable enhancement as they see fit.

Also while you're right that the CD timings on the real console have huge variations, it seems some game are strangely picky. I know the code in mednafen had to be tweaked several times because of games misbehaving.

bkoropoff:
@simias I'd be fine putting it behind a setting, maybe by turning the cache option into a `sync|async|precache` option.

Blocking the main thread produces noticeable audio/video stuttering on my ordinary 7200 RPM laptop hard disk.  Do I have a faulty setup, or are other people experiencing this as well?  If it's common, it might be better for the default to favor a smooth experience over determinism.  If it's not the default, it might be nice to display an OSD message when CD reading is slow so the user thinks to try changing the setting.

simias:
I play from a SSD using bin/cue and I haven't noticed any issue but obviously it's a rather optimal configuration. I think your modification is very much valuable, I just object to reducing accuracy by default. That could be just me though, maybe I'm being too nitpicky about that.

PR #244, opened 2017-10-15
CD access method cleanup

bkoropoff:
- Fix missing support for subchannel-only reads in the CCD backend
- Control sync/async/precache behavior through a unified setting, defaulting to sync to be conservative about emulator accuracy
- Detect slow CD reads and suggest the user try a different setting


bkoropoff:
cc @simias

PR #253, opened 2017-11-03
 System RAM for retro_get_memory_size 

celerizer:


rz5:
@celerizer - What does this PR enable/change in terms of behavior? Do e.g. the savestates work differently because of this or something?

celerizer:
retro_get_memory_data and retro_get_memory_size are needed for RetroAchievements to peek the console's memory. libretro.h says that they're supported but the actual functions were blank.

```
#define RETRO_ENVIRONMENT_SET_SUPPORT_ACHIEVEMENTS (42 | RETRO_ENVIRONMENT_EXPERIMENTAL)
                                           /* const bool * --
                                            * If true, the libretro implementation supports achievements
                                            * either via memory descriptors set with RETRO_ENVIRONMENT_SET_MEMORY_MAPS
                                            * or via retro_get_memory_data/retro_get_memory_size.
                                            *
                                            * This must be called before the first call to retro_run.
                                            */
```

The only other use is for cheats, which I'm guessing wasn't implemented because Mednafen already had a method.

Issue #184, opened 2017-04-15
Multidisk PBP and savestates [$25]


Kierek:
There is a weird behaviour regarding multidisk pbp and savestates. While in the middle of playing Parasite Eve I've been asked to change the disk. No problem, a couple of hotkeys later I'm playing again. 

The problem came out later, when I tried to load savestate made after disk change. It just doesn't work. Whenever I try to change location, it's unable to load it. I deducted that retroarch is still pointing to disk 1 and so I tried to change it. Nope, it looks like after loading savestate retroarch is losing count of available disk's and there's just "null" displayed. Using Disk Image Append didn't help either, I could choose disk index then, but I'm still experiencing the first problem.

Trying to change the disk to 2 before loading savestate doesn't help either. It seems that the only way for me to finish the game is by not using savestates. It's not a big problem for Parasite Eve, but I wanted to play some jrpgs after this and I just can't imagine doing it without savestates.

https://cloud.githubusercontent.com/assets/4080084/25064034/3a23e36c-21f2-11e7-894a-96e085e99d85.png

After a little bit of research it seems that the "null" problem happens only when saved on a disk different than the first one.

Kierek:
It works fine with several pbps and m3u file. After loading savestate from another disk it automatically changes retroarch's disk to said one and it works perfectly.

It's not a solution but at least a workaround.

daninthemix:
I was unable to replicate this problem on Resident Evil 2. It's an easy one to test as you can just switch to Disc 2 and begin a new game, which I did. I was able to relaunch Retroarch, run RE2 and load state immediately.

So either this is a problem with Parasite Eve, or a problem with your setup.

daninthemix:
I've added a bounty to this issue: https://www.bountysource.com/issues/44171556-multidisk-pbp-and-savestates

Kierek:
Thanks for confirming, I was worried that it was because of some corruption in my collection.

rz5:
@Kierek, @daninthemix,

A commit by @justinweiss was just merged and it purports to fix this bug. Please build it from source or use RA's core updater, double check if the core version ends with ae2f91f and check if it works.

ofry:
Please give me list of PS1 games which use more than 1 discs. Maybe I can try to test it.

daninthemix:
easiest one would be Resident Evil 2, as you can start on disk 2 immediately.

ofry:
[...]

PR #266, opened 2017-11-21
Fix save state loading with multi-disc PBPs

justinweiss:
I just ran into the problem mentioned in #184, and this is my stab at fixing it.

With this change, I can see the log messages adjusting the PBP image offset when loading savestates, which I didn't before. And I tested it by loading a Disc 2 savestate, and by restarting after loading the savestate and loading a Disc 2 save file. Both seemed to work.

There's only a few relevant lines, but I unfortunately had to move the function below the CDInsert / Eject functions, so it looks like there's more. 

I'm brand new to libretro, and rusty at C++, so I'd appreciate any feedback or suggestions!

### DIFF
--- a/libretro.cpp
+++ b/libretro.cpp
@@ -1883,59 +1883,6 @@ static void CloseGame(void)
    Cleanup();
 }
 
-int StateAction(StateMem *sm, int load, int data_only)
-{
-   SFORMAT StateRegs[] =
-   {
-      SFVAR(CD_TrayOpen),
-      SFVAR(CD_SelectedDisc),
-      SFARRAY(MainRAM.data8, 1024 * 2048),
-      SFARRAY32(SysControl.Regs, 9),
-      SFVAR(PSX_PRNG.lcgo),
-      SFVAR(PSX_PRNG.x),
-      SFVAR(PSX_PRNG.y),
-      SFVAR(PSX_PRNG.z),
-      SFVAR(PSX_PRNG.c),
-      SFEND
-   };
-
-
-   int ret = MDFNSS_StateAction(sm, load, data_only, StateRegs, "MAIN");
-
-   // Call SetDisc() BEFORE we load CDC state, since SetDisc() has emulation side effects.  We might want to clean this up in the future.
-   if(load)
-   {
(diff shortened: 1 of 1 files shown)


libretro.cpp
```diff
@@ -1993,6 +1940,70 @@ static void CDSelect(void)
...
+      SFEND
+   };
+
+
+   int ret = MDFNSS_StateAction(sm, load, data_only, StateRegs, "MAIN");
+
+   // Call SetDisc() BEFORE we load CDC state, since SetDisc() has emulation side effects.  We might want to clean this up in the future.
+   if(load)
+   {
+      if(CD_IsPBP)
```
justinweiss: This is the relevant section.

rz5:
@justinweiss Thanks for the commit, I'll merge it since the new behavior is exclusive to PBP files. Don't forget that the bug you referred to has a bounty associated, so once the participants in that thread acknowledge it as fixed, the bounty is yours.

Zapeth:
Well I'm not really familiar with the save state relevant code (specifically regarding the point in time when the according code is executed) so I can't really say much about that, but I guess it should be fine if it resolves the issue.

The only thing I noticed is that [code]CDC->SetDisc()[/code] gets called twice in the code, not sure if thats intended/necessary or not.
If not then you could remove either the CD*Eject() calls like below (and lose the "open/close" log messages)
[code]//CDEject();
CDC->SetDisc(CD_TrayOpen, (CD_SelectedDisc >= 0 && !CD_TrayOpen) ? (*cdifs)[0] : NULL,
         (CD_SelectedDisc >= 0 && !CD_TrayOpen) ? cdifs_scex_ids[0] : NULL);
//CDInsertEject();[/code]
or the CDC->SetDisc() call in between (again, not sure if it needs to be like this so testing would be recommended if its changed).

justinweiss:
Good catch. That wasn't intended -- the right fix would be to remove the `CDC->SetDisc()`. We need `(*cdifs)[0]->Eject(false)` to be called in order to move the PBP file to the right offset, and we also need `CD_TrayOpen` to be true for Eject to have any effect.  I could take a look tonight.

PR #277, opened 2017-12-17
Compilation fixes

albertofustinoni:


albertofustinoni:
The bug in `FileStream.cpp` is huge and will break the core if used with VFS

PR #296, opened 2018-01-13
Refactor SetDisc calls into function

remyabel:
The line was duplicated/long. Combined with the two ternary operators in the same call makes it hard to read. This minor patch should ease maintainability.


remyabel:
[Travis build](https://travis-ci.org/libretro/beetle-psx-libretro/jobs/328402569#L625) needs to restart.

PR #310, opened 2018-03-02
Fix for PGXP bugs

iCatButler:
- Pass clamped precise Z value to prevent divide by 0 on GPU (causes errors on Nvidia GPUs)
- Offset texture coordinates by 1/2 texel to avoid error on 1D look ups using Nvidia GPUs (issues #85)
- Prevent Vulkan renderer offsetting texcoord lookups by 1 during 1D lookups (caused error with above fix)


tabnk:
Nice

msmilkshake:
iCatButler I want to make you a question regarding pcsxr r. Is it possible to get in touch with you somewhere else?

Issue #350, opened 2018-06-24
[OpenGL] Castlevania Symphonia of the Night - Sprite Tiles and Distorted Sprites


Papermanzero:
Settings

Renderer: Hardware (OpenGL)
Software Framebuffer: Off
Adaptive Smoothing: Off
Internal GPU Resolution: 4x
Texture filtering: bilinear
Internal Color depth: 32bpp
Wireframe Mode: Off
Display Full VRAM: Off
PGXP operation mode: Memory only
PGXP vertex cache: On
PGXP perspective correct texturing: On
Widescreen mode hack: Off
Frame duping: Off
CPU Frequency Scaling: 100%
GTE Overclock: Off
GPU rasterizer overclock: 1x
Skip Bios: Off
Dithering pattern: 1x
Display interal FPS: Off
Initial scanline: 0
Last scanline: 239
Initial scanline PAL: 0
Last scanline PAL: 287
Crop Overscan: On
Additional Cropping: Off
Offset Cropped Image: Off
DualShock Analog button toggle: Off
Port 1 Multitap: Off
Port 2 Multitap: Off
Gun Cursoer: Cross
Mouse Sensitivity: 100%
CD Access Method: sync
Memcard 0 method: libretro
Enable memory card 1: on
Shared memcards: off
Increase cd loading speed: 2x

GPU: Nvidia 760 GTX

Description: With bilinear filter, Castlevania (PAL) has a tile effect (lines between sprites). With other filtering methods this is not appearing. Also PCSX1 with PGXP (standalone version) does not have this effect, with bilinear activated. Moreover some sprites like the candles are distorted. They are shaking left and right.

Screenshot (Tile effect):
![castlevania - symphony of the night-180624-210233](https://user-images.githubusercontent.com/10263536/41822973-4d9ce904-77f8-11e8-905a-4945fd93867d.png)

Screenshot (Distorted Sprites - See Candle):
![castlevania - symphony of the night-180624-210301](https://user-images.githubusercontent.com/10263536/41822976-590b4e70-77f8-11e8-8c4d-26869fb6bf0e.png)

rz5:
The distortion is most likely PGXP perspective correction related and that feature isn't really needed in a mostly 2D game.

The tiling effect bug is most likely caused by the 'texture filtering' core option being set to bilinear.

Papermanzero:
[...]

PR #365, opened 2018-07-01
Fix filtering and vertex cache errors in issue #350 and similar

iCatButler:
- Clamp texture coordinates to avoid sampling incorrect values when using filtering options
(replaces the need to offset texcoords by 1/2 pixel for perspective correction)
- Invalidate w component of cached vertices to selectively disable perspective correction when using them.


twinaphex:
Hi, do you still require my help on finding those entry points or did you already figure this out? Thanks a lot for the PR!

Issue #349, opened 2018-06-24
[OpenGL] Breath of Fire 3 - Sprites have black lines


Papermanzero:
Settings

Renderer: Hardware (OpenGL)
Software Framebuffer: Off
Adaptive Smoothing: Off
Internal GPU Resolution: 4x
Texture filtering: bilinear
Internal Color depth: 32bpp
Wireframe Mode: Off
Display Full VRAM: Off
PGXP operation mode: Memory only
PGXP vertex cache: On
PGXP perspective correct texturing: On
Widescreen mode hack: Off
Frame duping: Off
CPU Frequency Scaling: 100%
GTE Overclock: Off
GPU rasterizer overclock: 1x
Skip Bios: Off
Dithering pattern: 1x
Display interal FPS: Off
Initial scanline: 0
Last scanline: 239
Initial scanline PAL: 0
Last scanline PAL: 287
Crop Overscan: On
Additional Cropping: Off
Offset Cropped Image: Off
DualShock Analog button toggle: Off
Port 1 Multitap: Off
Port 2 Multitap: Off
Gun Cursoer: Cross
Mouse Sensitivity: 100%
CD Access Method: sync
Memcard 0 method: libretro
Enable memory card 1: on
Shared memcards: off
Increase cd loading speed: 2x

GPU: Nvidia 760 GTX

Description: All character sprites in Breath of Fire 3 (PAL) have vertical black lines.

Screenshot:
![breath of fire 3-180624-210050](https://user-images.githubusercontent.com/10263536/41822942-a45064a2-77f7-11e8-8d0b-c5619a33c979.png)

Papermanzero:
Also with commit  a8aa726, issue is still visible.

iCatButler:
I don't think the remaining lines are caused by texture filtering, they appear when using Nearest Neighbour sampling and even using the software renderer with resolution scaling.

Software with 4x scaling:
![breath of fire iii usa -180704-121504](https://user-images.githubusercontent.com/17948959/42274586-2479dc28-7f85-11e8-8149-5e5d00a64669.png)

It could be that the UV coordinates of the sprites/quads are actually set up to sample outside the correct area, but on original hardware and using the software renderer at native resolution this garbage data is never visible.

Papermanzero:
I guess following bug is caused by the same root cause (which is currently unknown).
https://github.com/libretro/beetle-psx-libretro/issues/351

iCatButler:
[...]

PR #370, opened 2018-07-16
Fix incorrect sampling on the edge of sprites. Fixes issue #349

iCatButler:
Borrow code from Vulkan implementation to fix incorrect sampling of texels on some sprite edges.
Replaces the UV min/max calculation code too.

### DIFF
--- a/rsx/rsx_lib_gl.cpp
+++ b/rsx/rsx_lib_gl.cpp
@@ -1687,6 +1687,140 @@ static bool retro_refresh_variables(GlRenderer *renderer)
    return reconfigure_frontend;
 }
 
+static void texCoord_preprocessing(
+		GlRenderer *renderer,
+		CommandVertex *vertices,
+		unsigned count)
+{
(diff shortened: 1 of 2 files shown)


twinaphex:
Thanks a lot for this!

twinaphex:
@iCatButler There are some regressions in some games -

see this in Ridge Racer Revolution for instance -

image

Hopefully you can find a solution for this since this is the only real regression I can find so far, the rest is an improvement overall over what went before.

twinaphex:
Another example (FF7) - notice the windows don't look correct -
image

iCatButler:
I've submitted a new PR that should fix these.

The problem was that the Vulkan renderer clamps texture coordinates immediately, and skips it for windowed textures. The OpenGL renderer clamps in the shader to handle taps from the various filtering options but wasn't setting valid bounds for these.

As the correct behaviour for windowed textures is to repeat within the window I'm just passing through the largest possible bounds so clamping does nothing and the coordinates will be truncated to fit the window instead.

Issue #306, opened 2018-02-17
Silent Hill White Squares with OpenGL


Papermanzero:
OpenGL seems to have several issues.
On the one hand the boot logo from the playstations is not correct (text is missing), on the other hand silent hill has a lot of graphical glitches.

Justa ctivate opengl, upscale the game and look at the results.

![silent hill-180217-231151](https://user-images.githubusercontent.com/10263536/36346110-3c7ac944-1438-11e8-84d7-deb0753f24ca.png)

simias:
It's the infamous mask bit emulation issue. It's well understood but it's a bit of a pain to implement.

twinaphex:
@simias We still badly require an implementation for this in the GL renderer.

07416:
Here is a screenshot of the game at native resolution:

![retroarch_2018_07_10_22_51_53_536](https://user-images.githubusercontent.com/14923168/42534322-21adee66-8495-11e8-9805-8b525f1b49dd.png)

07416:
Which renderer (video driver) should I use?

Papermanzero:
Vulkan or software

Papermanzero:
As simias already stated. It is the mask bit issue, which is described here in this news post:
https://www.libretro.com/index.php/introducing-vulkan-psx-renderer-for-beetlemednafen-psx/

An opengl implementation is possible, but it seems difficult. 
Nevertheless it is needed for the transparency effects in several games.

twinaphex:
This pull request should finally fix this issue -

https://github.com/libretro/beetle-psx-libretro/pull/377

Papermanzero:
This is fantastic. Bug is fixed. Meanwhile the core is really great. Time for a newspost on libretro.com ;-)

iiiGerardoiii:
@07416 hmm I don't see anything tbh

07416:
@iiiGerardoiii I didn't quickly find any editing setting to highlight the issue in Gimp, sorry,

![Uploading retroarch_2019_01_30_21_16_22_973.png…]()

Do you see now?

iiiGerardoiii:
oooh I see it now.

Good catch

07416:
> 
> 
> @iiiGerardoiii I didn't quickly find any editing setting to highlight the issue in Gimp, sorry,
> 
[...]

PR #377, opened 2018-07-22
Mask bits support using OpenGL stencil buffer

flyinghead:
Changed the way primitives are stored and the order in which they are drawn. Needed for
proper masking between opaque and semi-transparent primitives.
Probably not 100% correct but seems to fix the fog issue in Silent Hill (Issue #306)

### DIFF
--- a/rsx/rsx_lib_gl.cpp
+++ b/rsx/rsx_lib_gl.cpp
@@ -211,11 +211,17 @@ struct Framebuffer
    struct Texture _color_texture;
 };
 
-
-struct TransparencyIndex {
+struct PrimitiveBatch {
    SemiTransparencyMode transparency_mode;
-   unsigned last_index;
+	/* GL_TRIANGLES or GL_LINES */
    GLenum draw_mode;
+   bool opaque;
+   bool set_mask;
+   bool mask_test;
+   /* First index */
+   unsigned first;
+   /* Count of indices */
+   unsigned count;
 };
 
 template<typename T>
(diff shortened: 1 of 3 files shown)


flyinghead:
@twinaphex First shot at it.
Unfortunately it seems that most, if not all, other issues supposedly caused by lack of mask bits support are in fact unrelated.

twinaphex:
Thanks a lot, I will test it!

twinaphex:
The Silent Hill fogging is indeed fixed by this.

rsx/rsx_lib_gl.cpp
```diff
@@ -1834,12 +1796,22 @@ static void vertex_preprocessing(
...
       return;
 
    bool is_semi_transparent = v[0].semi_transparent == 1;
+   bool is_textured         = v[0].texture_blend_mode != 0;
+   /* Textured semi-transparent polys can contain opaque texels (when
+    * bit 15 of the color is set to 0). Therefore they're drawn twice,
+    * once for the opaque texels and once for the semi-transparent
+    * ones. Only untextured semi-transparent triangles don't need to be
+    * drawn as opaque. */
+   bool is_opaque = !is_semi_transparent || is_textured;
```
Themaister: Primitives which have mask test aren't necessarily opaque. They depend on the latest mask bit in the framebuffer, so it needs to be considered as blending.

rsx/rsx_lib_gl.cpp
```diff
@@ -990,126 +990,94 @@ static void GlRenderer_draw(GlRenderer *renderer)
...
-          * must be handled by the caller. This is because this command
-          * can be called several times on the same buffer (i.e. multiple
-          * draw calls between the prepare/finalize) */
-         glDrawElements(GL_TRIANGLES, opaque_triangle_len, GL_UNSIGNED_SHORT, opaque_triangle_indices);
-      }
+	  /* Mask bits */
+	  if (it->set_mask)
+		 glStencilOp(GL_KEEP, GL_KEEP, GL_REPLACE);
+	  else
+		 glStencilOp(GL_KEEP, GL_KEEP, GL_ZERO);
```
Themaister: I think this is kind of dangerous with textured, non-transparent primitives. You will end up filling the stencil buffer with either 0 or 1 depending on the alpha bit of the texture. You could use stencil_export, but that's not widely supported.

Themaister: Is the alpha in the framebuffer kept in sync with the stencil bit with texturing, copies, etc?

flyinghead:
It certainly doesn't cover all the cases and it's more of hack than anything else. The problem is I haven't been able to find another game than Silent Hill that uses mask bits. Supposedly Chrono Cross uses them during transition sequences but I haven't observed that. Same for Metal Gear Solid. Any pointer would be greatly appreciated.

I believe the correct solution would be to have stencil shaders and stencil-only passes, and may be some tricks here and there to keep stencil and alpha synchronized. I was about to go in that direction but it's a lot of engineering just for one game since this hack seems to work just fine.

rsx/rsx_lib_gl.cpp
flyinghead: Right, semi-transparent primitives with mask bits are also handled.

rsx/rsx_lib_gl.cpp
flyinghead: For textured opaque primitives, only the fully transparent texels are discarded (0x0000) by the shader and will not modify the stencil. So I think it's ok (?)

No, the stencil is not kept in sync for copies. However it is kept in sync (reset) in fill_rect. It *should* be kept in sync with primitives based on the gpu mask bits.

PR #404, opened 2018-08-11
add Vulkan texture filtering

hizzlekizzle:
the option currently requires a restart to show a change, and the bilinear-based shaders still have some transparency issues with which I was hoping @iCatButler could help me out.


twinaphex:
I'll currently merge this for now while we improve on this. Can you make a separate issue about the restart currently being necessary and ping iCatButler in it?

iCatButler:
@hizzlekizzle I've built the latest code and it's looking pretty great. I couldn't see any obvious problems with transparencies specifically, if you can point them out I'll take a look.

The only thing that stands out to me is that the sampling code still needs to clamp UVs to the limits of the primitive (it's sampling over the edges still). In the OpenGL code those are passed through from via vertex attributes. If you'd like I can look at fixing that up the same way it's done in OpenGL.

hizzlekizzle:
>The only thing that stands out to me is that the sampling code still needs to clamp UVs to the limits of the primitive (it's sampling over the edges still). In the OpenGL code those are passed through from via vertex attributes. If you'd like I can look at fixing that up the same way it's done in OpenGL.

Ah, yes, the clamping is what I was referring to. The opacity stuff seems to be doing its job. I would love it if you could take a look at doing that clamping in the vulkan renderer, as well! :)

Issue #407, opened 2018-08-13
[Vulkan/Beetle PSX HW] RE2 - Dual Shock Ver. - Texture warping error on Leon's face


cheds89:
Hi,

This is another visual bug I've noticed with Resident Evil 2 - Dual Shock Ver. with the Beetle PSX HW core using the Vulkan driver -

![resident evil 2 - dual shock ver usa -180813-195024](https://user-images.githubusercontent.com/42318042/44025556-cff7bf66-9f33-11e8-9949-8753d8fb6fb6.png)

The texture warping on Leon's face, particularly the eye area, seems really wrong to me. 

A comparison with the same scene using the Beetle PSX core (Software) shows how it should look -

![resident evil 2 - dual shock ver usa -180813-194023](https://user-images.githubusercontent.com/42318042/44025647-068b4e8a-9f34-11e8-90c3-2ff7732862b0.png)

Screenshots are taken at 4x GPU Internal Resolution.

I hope these are helpful.

Thanks!

twinaphex:
Why is everything blue in your screenshot though? It makes it a bit hard to make any kind of comparisons.

Also, I think the texture warping on his face is just as a result of PGXP being disabled. The PlayStation suffers from lack of integer texture coordinates, so that and the lack of perspective correct texturing makes it that textures can warp. Add to that lack of subpixel precision and you have polygons too that wobble.

What you can try is setting PGXP Internal mode to ' Memory only'. With Resident Evil 1 this greatly helps with the facial warping/distortion.

So basically, I don't think this is an emulator bug, you are just running into one of the limitations of the PSX hardware. Luckily, we have additional features in the emulator that can mitigate these issues quite successfully nowadays, so try PGXP out.

hizzlekizzle:
The red and blue channels are swapped on vulkan screenshots (on some GPUs but not others, maybe??)

Here's what it looks like with those channels swapped back in photoshop:
![untitled-1](https://user-images.githubusercontent.com/2805586/44033754-f842242e-9ed0-11e8-9b01-53787aac5dad.png)

cheds89:
Thanks for both your replies.

Yep, the screenshots seem to have their blue and red channels reversed when using the Vulkan driver. Thanks for fixing hizzlekizzle. I also found it mentioned here - 

https://forums.libretro.com/t/red-blue-color-shift-in-screenshots-taken-with-vulkan-driver/10565

[...]

Issue #385, opened 2018-07-26
[OpenGL/Vulkan/Upscaled Software] Soul Blade/Edge - Water does not render correctly


saftle:
Original reported by @twinaphex here: https://github.com/libretro/beetle-psx-libretro/issues/43#issuecomment-220839628

> Both software rendering at higher than 1x resolution and the GL rendering produce the same graphics errors here. Only 1x software mode renders the water correctly. Top is GL, middle is software renderer at 4x, and bottom is the software renderer at 1x.
> 
> https://cloud.githubusercontent.com/assets/1075274/15454927/e5a59762-2045-11e6-9dcf-710af002807d.png
> 
> https://cloud.githubusercontent.com/assets/1075274/15454928/e5a6abe8-2045-11e6-8763-96ed56cd7105.png
> 
> https://cloud.githubusercontent.com/assets/1075274/15454929/e5a731bc-2045-11e6-963e-375a9911deda.png

I was also able to confirm all of what @twinaphex posted against the newest commit using default settings: 581a08fc62a21515e4f9d9fb2d0e96e5da434f43

The problem exists also when using the Vulkan renderer, and not just the GL renderer, or the software renderer when upscaled.

I have also attached a save state for easy debugging: 
[soul blade (usa) (v1.1).state.zip](https://github.com/libretro/beetle-psx-libretro/files/2232746/soul.blade.usa.v1.1.state.zip)

saftle:
@iCatButler Ah I see, so then it also affects Final Doom, Hexen, and Star Wars Dark Forces as mentioned by @twinaphex here: https://github.com/libretro/beetle-psx-libretro/issues/43#issuecomment-220841693

twinaphex:
@iCatButler Could you attempt making a diff patch that implements this? (looking for triangles that would be a single pixel high and converting them to quads that fill out the whole line at whatever scale) Once I have a diff that works for the problematic games (Soul Blade/Final Doom/Doom/Hexen), I could then look at figuring out a way we could  integrate this in the emulator, maybe on a per-game basis.

iCatButler:
@twinaphex okay, I've implemented a hack that fixes the rendering in these games:
https://github.com/iCatButler/beetle-psx-libretro/commit/e6b22a114624b72d3f8e5722b3c3cf3a38b6fabc.patch

[...]

PR #411, opened 2018-08-28
[squash and merge] Create core option for line-to-quad hack

hizzlekizzle:


twinaphex:
Hey @iCatButler , I asked @hunterk to take alook into implementing this so here it should be. Sorry for the delay so far!

twinaphex:
And thanks to @hizzlekizzle for implementing this!

Unrul3r:
What does this option do?

hizzlekizzle:
It tunes a heuristic that detects single-line polygons and expands them into a quad instead. The aggressive mode fixes some games that were previously unplayable, like Duke Nukem and Dark Forces and fixes some effects in other games. However, it can also cause false positives that lead to funky geometry where it doesn't belong.

Unrul3r:
@hizzlekizzle I see. Thank you for the explanation! How would I detect if a game I play needs this or not (apart from the unplayability like the ones that you've mentioned)?

hizzlekizzle:
@Unrul3r tbh, I don't really know of any way other than just being familiar with the game(s).

Jj0YzL5nvJ:
Didactic documentation of this hack, pros #385 and cons #407.

Issue #398, opened 2018-07-31
]


CsoiretgeFsM:

iCatButler:
I suspect that is related to this issue in Breath of Fire 3: https://github.com/libretro/beetle-psx-libretro/issues/349#issuecomment-402451168

There's code that originates in the Vulkan renderer that fixes the issue and was added to the OpenGL implementation, so that should be fixed in the latest build.
The best solution is probably to move that code to before any of the renderers so that it works for all of them without duplicating the same code in each of them. I'll look into doing that when I get a chance.

CsoiretgeFsM:
Thanks for all your hard work iCB. As an end user I'm sorry all I can contribute is bug and playtesting. If I had taken my coding classes more seriously maybe I could've helped speed things up here but alas.

twinaphex:
> The best solution is probably to move that code to before any of the renderers so that it works for all of them without duplicating the same code in each of them. I'll look into doing that when I get a chance.

I very much approve of this idea, it could remove logic out of the individual GL/Vulkan renderers and as a nice side effect also fix this issue for the software renderer. Would you like to take a stab at it?

twinaphex:
I just merged @iCatButler's PR which is supposed to solve this issue. Can the OP test this again using the latest updated core from the buildbot?

orbea:
@CsoiretgeFsM Can you please test this?

PR #448, opened 2018-11-25
Fix PGXP "+CPU" integration

iCatButler:
Add PGXP calls for instructions that were previously missed. Should improve support for games like Croc, MGS, Alundra 2 and others that rely on CPU instructions for vertex handling.

### DIFF
--- a/mednafen/psx/cpu.cpp
+++ b/mednafen/psx/cpu.cpp
@@ -2131,6 +2131,9 @@ pscpu_timestamp_t PS_CPU::RunReal(pscpu_timestamp_t timestamp_in)
 	uint32 result = GPR[rs] - GPR[rt];
 	bool ep = (((GPR[rs] ^ GPR[rt])) & (GPR[rs] ^ result)) & 0x80000000;
 
+	if (PGXP_GetModes() & PGXP_MODE_CPU)
+		PGXP_CPU_SUB(instr, result, GPR[rs], GPR[rt]);
+
 	DO_LDS();
 
 	if(MDFN_UNLIKELY(ep))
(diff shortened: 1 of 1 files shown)


twinaphex:
Hey, that's cool! I will be checking out Metal Gear Solid later today again and maybe showing a new video for it on Youtube.

twinaphex:
BTW @iCatButler, it would still be great if you could do this -

https://github.com/libretro/beetle-psx-libretro/issues/398#issuecomment-410935683

I unfortunately don't think I would be able to do that myself. I took a few stabs at it before but got nowhere.

iCatButler:
@twinaphex I've taken a few shots at it myself from various directions. I'm beginning to think it might actually be simpler to implement a specific software renderer solution rather than share the GL/Vulkan code.

So far as I can tell the error only occurs on quad primitives rather than sprites in software, which are handled separately, whereas these eventually use the same path in the hardware renderers and so both get the fix applied.

twinaphex:
Here is a glitch I have noticed so far with PGXP + CPU - this is with Wipeout XL -

image

This is how it should look like instead -

image

iCatButler:
I think I've got the software renderer sampling error fixed now. There was already a sub-texel offset applied to all the base UVs which was causing it to sample out of bounds on 2D quads regardless of the UV orientation (in hardware the error was only when the UVs were flipped).

I've changed it so this offset is shrunk as the scaling factor increases, which solves the problem for positive UV/XY derivatives and allows us to use the same code from the hardware renderers to fix flipped UVs. I've currently got it sharing the code for both GL and software renderers. Hopefully it'll be pretty quick to add integrate with the Vulkan renderer and I can get it checked in :)

PR #449, opened 2018-11-26
vulkan: Remove assert.

orbea:
Warning: Please review and test this, I am not sure that its correct, only that it boots games with amdgpu and vulkan now.
```
Vulkan error at parallel-psx/vulkan/semaphore.hpp:40.
terminate called without an active exception
```


twinaphex:
Ah, I think it is probably fine to remove this.

Issue #445, opened 2018-11-14
[Vulkan] Final Fantasy 8 Hardware Core becomes unstable, crashes after entering room


xeropresence:
# First and foremost consider this:
Bug is not reproducible with mednafen, or the software renderer.
    
## Description

When entering the train compartment containing zell after leaving balamb to dollot the hardware renderer stops working correctly. Attempting to load a a save state results in corrupted textures. While the display stops rendering, the gameplay does sound like its continuing on in the background.
### Expected behavior

[What you expected to happen]

Graphics to continue running properly

[What is actually happening]

ff8 beetle vulkan hw render
https://streamable.com/twrfj
ff8 beetle ogl hw render
https://streamable.com/vsla4
ff8 beetle sw render
https://streamable.com/nj488
ff8 mednafen 
https://streamable.com/6w7zz

### Steps to reproduce the bug

[Final Fantasy VIII Disc1.1.zip](https://github.com/libretro/beetle-psx-libretro/files/2580066/Final.Fantasy.VIII.Disc1.1.zip)

Load save game
Move into next compartment
Move into next compartment

### Bisect Results

[Try to bisect and tell us when this started happening]

### Version/Commit
[If you're using RetroArch, after the core has started, this information is on the bottom left]

- Beetle PSX: RetroArch Beetle PSX 0.9.44.1 74bd755

### Core options
No core options were changed for the recorded videos

### Environment information

- OS: Windows 10 1809
- GPU: 1080 ti 416.34

xeropresence:
Interestingly , if I changed the cpu frequency scaling to ~70% the issue does not occur.

rz5:
I can confirm that graphical glitches do happen on the vulkan renderer.

iCatButler:
The same bug occurs in the starting forest scene of Chrono Cross' attract mode.

It appears to have been introduced when the vertex size was increased to include UV limits (to clamp sampling to the edges of primitives). If I remove some of those new member variables, making the vertices smaller, or increase the block size of the Vulkan Device's PerFrame vbo_chain allocator then the glitching doesn't occur.

It may have been an existing error that went undetected and it's writing over the end of the buffer now that the vertices are larger.

theoldsport:
@iCatButler related to #422? Vertex explosions tending to occur on Vulkan, generally when cpu overclocking enabled

iCatButler:
[...]

Issue #422, opened 2018-09-04
[VULKAN] Chrono Cross severe graphical glitches with overclocking


Electricstar:
![hot garbage](https://user-images.githubusercontent.com/14850864/45060957-585b3c00-b070-11e8-8aa3-e811e0dc5139.png)

Chrono Cross has severe glitches when overclocking, most of which start happening at 150% CPU speed. However, in some situations, having the clock set to even 110% will cause graphical glitches. For some reason though, changing the resolution will undo the glitching, at least until you go somewhere problematic again (in terms of having high overclock).

Electricstar:
'Running away from battle' test @ 240% OC

OpenGL:
![cchardwareogl](https://user-images.githubusercontent.com/14850864/47045251-594bb700-d160-11e8-9520-47f59e7f6d9d.png)

![cchardwarevulkan](https://user-images.githubusercontent.com/14850864/47045276-6f597780-d160-11e8-904a-ad516082e563.png)

theoldsport:
This happens in Silent Hill and some other games too, exact behavior as you described. Perhaps change the title to be more general?

iCatButler:
@Electricstar there should be a fix in the latest build for this if you could test it.

theoldsport:
@iCatButler no more vertex explosions with Vulkan in Silent Hill, or any of the games I saw similar problems. Deep down I hoped you might inadvertently solve this issue #368, but it's clearly unrelated.

PR #455, opened 2018-12-04
Fix chain allocator behaviour

iCatButler:
- Originally had a single host pointer to the last block allocated. If more than one block was allocated when the allocator was reset using discard() this caused a mismatch between the block allocated from and the one written to.
- Now stores a pointer to the data in each block and returns the correct one when allocating.

Should fix issues #445 and #422

Also changed the vertex declaration for UV clamping values to be single channel floats (doesn't appear to have been causing errors, but seems more correct).


rz5:
I just tried to replicate bug #445 with your changes, it does seem to be fixed. In particular, I tried replicating the bug while having the CPU overclock core option set to 500%. The bug still didn't happen.

So good job! Seems like you're getting the hang of the Vulkan code.

twinaphex:
Awesome work! Maybe this will also fix the graphics glitches I got with Jumping Flash when overclocking too high.

iCatButler:
@twinaphex it might be worth making TinyTiger aware of this if the Vulkan code is being used in any other projects.

PR #488, opened 2019-02-18
Silence fprintfs coming from the core

rdanbrook:
If we really want to see this information, it should be processed in the frontend with the proper retro_log_level rather than the core writing directly to stderr.


orbea:
It would probably be better to fix them correctly instead of just commenting them, I suspect they will never be fixed if just commented..

If they are not useful then it would be better to outright remove them.

rdanbrook:
I'll redo this later with the problem fixed in the libretro portion.

PR #534, opened 2019-09-05
Update core options

jdgleaver:
This PR makes the following changes to the core options:

- Update `libretro_core_options` to v1.3 format

- Complete overhaul of option sublabels

- Irrelevant options are now hidden automatically depending upon current renderer

- General tidy up of everything :)

### DIFF
--- a/libretro-common/include/libretro.h
+++ b/libretro-common/include/libretro.h
@@ -1117,7 +1117,7 @@ enum retro_mod
                                             * This may be still be done regardless of the core options
                                             * interface version.
                                             *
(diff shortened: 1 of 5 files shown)


Awakened0:
```
      "Widescreen Mode Hack",
      "When enabled, forces content to be rendered with an aspect ration of 16:9.
```
Aspect ration should be aspect ratio. I always make that typo too.

jdgleaver:
@Awakened0 Thanks - good catch!

Fixed by PR #535  :)

Awakened0:
No prob, thanks!

Issue #510, opened 2019-06-02
[Beetle PSX HW] Pixel aspect ratio should be 8:7 not 5:4 in 256px width mode


ggdrt:
## Description

Currently by default the core stretches 256px width framebuffers by 25% to a 320px width 4:3 display aspect ratio, which results in an incorrect 5:4 pixel aspect ratio. The correct pixel aspect ratio is 8:7 (~14.286% stretch), which all currently visible Mednafen releases on https://mednafen.github.io/releases/ (0.9.48 through 1.22.2) handle correctly. For a 256px width framebuffer this 8:7 PAR results in around a 1.22:1 DAR with 240 lines (alternatively 1.306:1 DAR with vertical overscan cropping to 224 lines). The missing width for the full 4:3 DAR should be horizontal overscan area, as shown in the image below. 

### Image of issue

The upper image demonstrates the current behavior on Beetle PSX HW while the lower image is the same title running on Mednafen 1.22.2 with expected behavior. The 2D character sprites and the portrait image on Beetle PSX HW are unnaturally squat.

![Image](https://i.imgur.com/TPCsvAL.png)

### Testing results

This has been an issue in the HW renderer for as far back as I can tell, and isn't handled correctly without the user manually setting the aspect ratio through the frontend. The issue seems to be irrespective of core options or Windows platform distributions (7 and 10 ver1709/1803/1809 tested). ~~#5 also brings up a similar issue four years ago, although the user's claims that the attempted manual solution resulted in the correct resolution are most definitely incorrect based on the calculations provided below.~~ (In the linked issue, software renderer was actually performing correct behavior.)  In any case, I'm not sure it would make sense to have end users use the RetroArch frontend to manually solve inaccurate behavior on the core side. 

I haven't had the time to verify scaling behavior for other width modes yet, but I wouldn't be surprised if the core also handles those incorrectly.

### Info dump

A lot of this information was strewn over the web in various places so I wanted to compile it together for thoroughness.

The sampling rate on an NTSC analog television for 1:1 PAR at 480i is 135/11 MHz. This sampling rate drops to 135/22 MHz for 1:1 PAR at 240p. The PS1 gpu is 53.69 MHz with a dotclock divider of 10 for 256px mode, resulting in pixels outputted at 5.369 MHz. So relative to this 135/22 MHz 1:1 PAR baseline, the PAR of the PS1 at 256px width is (135/22 MHz)/(5.369) ~= 1.1429 ~= 8/7. 

[...]

PR #538, opened 2019-09-27
Update core options sublabels

ggdrt:
This PR does the following:

1. Updates a few core option sublabels to be more descriptive, such as certain options not working with the hardware renderer at the moment
2. Removes some erroneous information regarding horizontal overscan cropping and rewrites that option sublabel
3. Fixes and clarifies overscan cropping comment in libretro.cpp (hopefully to aid in fixing a crop_overscan alignment issue; I'll open an issue for this when I have some more specifics I can point to)

Will follow up by updating the core's libretro doc page with pictures and better descriptions and open an issue for the aforementioned crop_overscan option in the near future.

### DIFF
--- a/libretro_core_options.h
+++ b/libretro_core_options.h
@@ -55,7 +55,7 @@ struct retro_core_option_definition option_defs_us[] = {
    {
       BEETLE_OPT(renderer),
       "Renderer (Restart)",
-      "Choose video renderer. 'Software' is the most accurate but has the highest performance requirements when running at increased internal GPU resolutions. 'Hardware' selects automatically the 'OpenGL' or 'Vulkan' renderer, depending upon the current RetroArch video driver setting. While less accurate, these renderers improve performance and enable various enhancements such as texture filtering and perspective correction. Core will automatically fall back to the software renderer if 'Hardware' is selected but the RetroArch video driver is not set to Vulkan or an OpenGL 3.3-compatible driver.",
+      "Choose video renderer. 'Software' is the most accurate but has the highest performance requirements when running at increased internal GPU resolutions. 'Hardware' automatically selects the 'Vulkan' or 'OpenGL' renderer depending upon the current libretro frontend video driver. If the provided video driver is not Vulkan or OpenGL 3.3-compatible then the core will fall back to the software renderer. While less accurate, the hardware renderers improve performance and enable various enhancements such as texture filtering and perspective correction.",
       {
          { "hardware", "Hardware" },
          { "software", "Software" },
@@ -405,7 +405,7 @@ struct retro_core_option_definition option_defs_us[] = {
    {
       BEETLE_OPT(image_offset),
       "Offset Cropped Image",
-      "When 'Crop Overscan' is enabled, allows the resultant cropped image to be offset horizontally to the right by the specified number of pixels. May be used to correct alignment issues. As with 'Crop Overscan', currently only supported by the software renderer.",
+      "When 'Crop Overscan' is enabled, allows the resultant cropped image to be offset horizontally to the right (positive) or left (negative) by the specified number of pixels. May be used to correct alignment issues. As with 'Crop Overscan', currently only supported by the software renderer.",
       {
          { "disabled", NULL },
          { "-4 px",    NULL },
--- a/libretro_core_options.h
+++ b/libretro_core_options.h
@@ -55,7 +55,7 @@ struct retro_core_option_definition option_defs_us[] = {
    {
       BEETLE_OPT(renderer),
       "Renderer (Restart)",
-      "Choose video renderer. 'Software' is the most accurate but has the highest performance requirements when running at increased internal GPU resolutions. 'Hardware' selects automatically the 'OpenGL' or 'Vulkan' renderer, depending upon the current RetroArch video driver setting. While less accurate, these renderers improve performance and enable various enhancements such as texture filtering and perspective correction.",
+      "Choose video renderer. 'Software' is the most accurate but has the highest performance requirements when running at increased internal GPU resolutions. 'Hardware' selects automatically the 'OpenGL' or 'Vulkan' renderer, depending upon the current RetroArch video driver setting. While less accurate, these renderers improve performance and enable various enhancements such as texture filtering and perspective correction. Core will automatically fall back to the software renderer if 'Hardware' is selected but the RetroArch video driver is not set to Vulkan or an OpenGL 3.3-compatible driver.",
       {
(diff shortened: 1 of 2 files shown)


libretro_core_options.h
```diff
@@ -55,7 +55,7 @@ struct retro_core_option_definition option_defs_us[] = {
    {
       BEETLE_OPT(renderer),
       "Renderer (Restart)",
-      "Choose video renderer. 'Software' is the most accurate but has the highest performance requirements when running at increased internal GPU resolutions. 'Hardware' selects automatically the 'OpenGL' or 'Vulkan' renderer, depending upon the current RetroArch video driver setting. While less accurate, these renderers improve performance and enable various enhancements such as texture filtering and perspective correction.",
+      "Choose video renderer. 'Software' is the most accurate but has the highest performance requirements when running at increased internal GPU resolutions. 'Hardware' selects automatically the 'OpenGL' or 'Vulkan' renderer, depending upon the current RetroArch video driver setting. While less accurate, these renderers improve performance and enable various enhancements such as texture filtering and perspective correction. Core will automatically fall back to the software renderer if 'Hardware' is selected but the RetroArch video driver is not set to Vulkan or an OpenGL 3.3-compatible driver.",
```
rz5: Remember, libretro cores aren't exclusive to RetroArch. Consider changing this sublabel to something that is generic for any libretro frontend.

libretro_core_options.h
```diff
@@ -405,7 +405,7 @@ struct retro_core_option_definition option_defs_us[] = {
    {
       BEETLE_OPT(image_offset),
       "Offset Cropped Image",
-      "When 'Crop Overscan' is enabled, allows the resultant cropped image to be offset horizontally by the specified number of pixels. May be used to correct alignment issues.",
+      "When 'Crop Overscan' is enabled, allows the resultant cropped image to be offset horizontally to the right by the specified number of pixels. May be used to correct alignment issues. As with 'Crop Overscan', currently only supported by the software renderer.",
```
rz5: 'to the right' - This is wrong. The offset can be positive (to the right) or negative (to the left).

rz5:
@ggdrt - Thanks for giving this a look. I left some comments in some lines on your changes, please review them.

Also, don't be afraid to shorten the sublabels either. If you can make these descriptions more succint while still being clear, I encourage it.

libretro_core_options.h
ggdrt: That's true; I was basing it off the original text that already mentioned RetroArch. I'll change it to something more neutral.

libretro_core_options.h
ggdrt: The reason why I put 'to the right' is because the original description doesn't specify if a positive offset is shifting the image to the right or to the left. I assumed that adding 'to the right' would make it more obvious that a positive shift would be to the right and a negative shift 'to the right' would move the image left.

Maybe something like "When 'Crop Overscan' is enabled, allows the resultant cropped image to be offset horizontally to the right (if offset is positive) or to the left (if number of pixels is negative) . . ." would be less ambiguous?

ggdrt:
@rz5 Thanks for the comments. I'm still a bit iffy on how to word the image_offset option. I'll get back to this later in the day when I'm free.

One of the biggest things right now is that the sublabels and descriptions in the docs are out of sync. I think I'll have a clearer idea of how to do that after updating the docs.

ggdrt:
Should be good to go now.

ggdrt:
Hrm, wasn't aware of that, but that's good to know. Might be too technical for the sublabels but I'll try putting that into the docs for now.

jdgleaver:
@ggdrt I just noticed this PR - good stuff, but I have a couple of comments:

- For the `Crop Overscan` option, you removed the warning about incorrect aspect ratios. I thought this was rather important!

- If the `Crop Overscan`, `Additional Cropping` and `Offset Cropped Image` options only apply when using the software renderer, they should be hidden when using a hardware renderer. This is easy - you just need to add them to this case statement:

https://github.com/libretro/beetle-psx-libretro/blob/e4c36745a35a492b947bbc5ee92a3d4f120b69db/libretro.cpp#L3520-L3585

ggdrt:
@jdgleaver I removed the warning because I'm fairly certain it's incorrect. If the cropping is removing the pillarboxes and stretching the remaining image to 4:3, I believe it's overstretching. You can double-check my math at #510. This doesn't just apply to 256px width -- all of the other width modes have this same issue as well. Upstream mednafen preserves the original pixel aspect ratios by cropping the bars and resizing the display window's width but for libretro frontends it seems you need to manually set a custom viewport to maintain the correct pixel aspect ratio. This messed up display aspect ratio is another related issue to why the hardware renderer isn't working properly for people using RetroArch with CRTs. (Also as far as I remember, none of the other official Sony emulators stretch out to a full 4:3 either from when I checked a couple of months ago; they seem to maintain somewhere between a 1.22~1.29 aspect ratio with non-visible horizontal overscan)

I have those changes for hiding those options in a local branch right now but I didn't want to commit them yet since I'm not sure just exactly how those options interact with the hardware renderers. Empirically testing they don't work, but whether that's due to a bug or because the features just aren't implemented I can't say for sure.

jdgleaver:
> @jdgleaver I removed the warning because I'm fairly certain it's incorrect. If the cropping is removing the pillarboxes and stretching the remaining image to 4:3, I believe it's overstretching. You can double-check my math at #510. This doesn't just apply to 256px width -- all of the other width modes have this same issue as well. Upstream mednafen preserves the original pixel aspect ratios by cropping the bars and resizing the display window's width but for libretro frontends it seems you need to manually set a custom viewport to maintain the correct pixel aspect ratio. This messed up display aspect ratio is another related issue to why the hardware renderer isn't working properly for people using RetroArch with CRTs. (Also as far as I remember, none of the other official Sony emulators stretch out to a full 4:3 either from when I checked a couple of months ago; they seem to maintain somewhere between a 1.22~1.29 aspect ratio with non-visible horizontal overscan)

Okay - I am astonished!

Turns out this was a critical memory failure on my part. I always remembered PSX games as having  a 4:3 aspect ratio. But I just got my PSX out of storage - and on real hardware, you're absolutely correct.

I've been wrong all this time...

Anyway, sure - your core description is fine :)

> I have those changes for hiding those options in a local branch right now but I didn't want to commit them yet since I'm not sure just exactly how those options interact with the hardware renderers. Empirically testing they don't work, but whether that's due to a bug or because the features just aren't implemented I can't say for sure.

Ah, okay - best not to hide the options then, until we're sure what's going on here.

Issue #543, opened 2019-10-10
Display area misaligned by default when Crop Overscan is enabled for SW renderer


ggdrt:
## Description

Display area is misaligned by default when Crop Overscan option is enabled in software renderer. It crops too few pixels on one side and too many pixels on the other side. Issue occurs regardless of internal resolution, window scale, or frontend.

Issue doesn't apply for the hardware renderers since Crop Overscan option doesn't work for them.

### Expected behavior

Crops properly to the display area by default (basically for any game that sets GP1(06h) X1=260h and X2=C60h, which should be most of them) for all the possible width modes. This should be expected behavior without having to utilize the Offset Cropped Image option.

### Actual behavior

Some examples:

[256px width mode, crop overscan off](https://user-images.githubusercontent.com/45282415/66547181-845ab380-eaf3-11e9-934f-1888d320a4c0.png)
[256px width mode, crop overscan on](https://user-images.githubusercontent.com/45282415/66547165-7d33a580-eaf3-11e9-818c-ddfb8e015741.png)

*Display area is offset approx 2 pixels to the right by default in 256px mode.*

[320px width mode, crop overscan off](https://user-images.githubusercontent.com/45282415/66550774-f256a900-eafa-11e9-8639-a3f1b3cc3dd9.png)
[320px width mode, crop overscan on](https://user-images.githubusercontent.com/45282415/66550783-f682c680-eafa-11e9-97a3-c81903707306.png)

*Display area is shifted approx 1 pixel to the right by default in 320px mode.*

[512px width mode, crop overscan off](https://user-images.githubusercontent.com/45282415/66561507-79614c80-eb0e-11e9-9a80-454e6961af44.png)
[512px width mode, crop overscan on](https://user-images.githubusercontent.com/45282415/66561515-7e260080-eb0e-11e9-8e1f-417685d32eb8.png)

*Display area is shifted approx 3 pixels to the left by default in 512px mode.*

Haven't tested with other width modes yet.

### Thoughts

Adjusting the hardcoded shifts for each width mode (e.g. 10 to 12 for padded 280, 14 to 15 for padded 350, etc.) in the lines below seem to give the correct results. I wasn't able to figure out how the current offset values were derived.

https://github.com/libretro/beetle-psx-libretro/blob/9f1341a9c3f881b1a552d0c0c4e8d3e22b2ed578/libretro.cpp#L3864-L3904

[...]

PR #550, opened 2019-10-17
Adjust outdated overscan cropping code

ggdrt:
Fixes issue #543.

The 400px -> 366px crop is for the 368px display mode. Some 368px width games are a little finicky to deal with -- sometimes using 368px results in undercropping while sometimes using 364px results in overcropping. 366px was chosen as a compromise and it seems to work for the most part. Left a comment in the code in case this needs to be adjusted later.


rz5:
Good stuff ggdrt

Issue #525, opened 2019-07-23
Odd geometry behavior between SW & HW renderer on games with built in letterboxing


klepp0906:
# First and foremost consider this:
- This is a DEVELOPER oriented channel dedicated to report bugs and suggest features about the libretro core, please do not ask for help;
- The upstream bug tracker is a forum thread, located at https://forum.fobby.net/index.php?t=msg&th=1114;
    - Note: As of October 2018, Microsoft email providers (think email addresses ending with hotmail/live/outlook) are rejecting email from the forum. If you want to register an account in the upstream project's forum, consider using another mail provider.
- For game-related bugs:
    1) First, check if it's already reported on the upstream bug tracker;
    2) If it's not reported there, try to reproduce the bug with mednafen;
    3) If it's reproducible with mednafen, report it on the upstream bugtracker;
    4) If it's not reproducible with mednafen, ONLY THEN should you report it here.
    
## Description
Games that have letterboxing across seem to function properly in the SW core, however in the HW core, it seems it only letterboxes the top - and leaves the bottom untouched. 

Here are two shots of a game that doesnt have letterboxing with identical settings.

SW renderer - https://www.dropbox.com/s/41wvpwgkg9187tr/sw.jpg?dl=0
HW renderer - https://www.dropbox.com/s/3zvz3oo5ih4odds/hw.jpg?dl=0

Result is identical.

Here are two shots of a game that does have built-in letterboxing for lack of a better word.  Identical settings to the above game/shots, and to eacthother.

SW renderer -  https://www.dropbox.com/s/8lieoyxypepx196/sw2.jpg?dl=0
HW renderer - https://www.dropbox.com/s/tooa4dl4ogxqa0w/hw2.jpg?dl=0

May be hard to see with the scaled down images, but notice how the SW renderer has both "letterboxes" and theyre straight across (barring the slight curve from the shader).

The HW renderer seems to be skipping the bottom black bar alltogether, and leaving the default curvature+corner behavior from non letterboxed games, while treating the top as expected.

### Expected behavior

For the HW renderer to behave like the SW renderer - and for the bottom letterbox to be cut off like the top.   Even if i were to reposition the Y axis upward, the geometry of the top and the bottom would be different.  (not to mention the loss of convenience by being able to potentially use either renderer without major adjustments to viewport).

### Steps to reproduce the bug

1. Applying an overlay may be helpful for position relative to the space. 
[...]

Issue #495, opened 2019-03-20
Beetle PSX Hardware Rendering Stretches Video Which Breaks Interlacing


xchimp:
## Description

When using hardware rendering, the screen stretches and breaks an interlacing shader when using a CRT monitor at 3840x480.

### Expected behavior

The video to be it's proper aspect ratio depending on the game.
![software](https://user-images.githubusercontent.com/31327707/54671246-8875a400-4ac3-11e9-9fec-9fd2b3c96add.jpg)

### Actual behavior

The video is being stretched, creating excessive black bars.
![hardware](https://user-images.githubusercontent.com/31327707/54671495-0c2f9080-4ac4-11e9-977b-5a9b1c310958.jpg)

### Steps to reproduce the bug

1. Set renderer to hardware
2. Enable interlacing shader

### Version/Commit

- Beetle PSX: 0.9.44.1 5359198

### Core options
Renderer = hardware

### Environment information

- OS: Windows 10 x64
- GPU: R9 380 4GB | Driver version: 19.3.1

hizzlekizzle:
Does it act any better with the software renderer? Do you have integer scaling enabled?

xchimp:
Yes, the expected behavior is a picture of it running in software mode, I should have clarified. Yes, it's enabled, disabling it doesn't fix it.

hizzlekizzle:
Are you increasing the internal resolution? If so, I believe this is acting as expected if not as intended/hoped. That is, if you increase the internal resolution to be larger than the actual screen/res, RetroArch will downsample it to fit the screen, which does not take integer scaling into account.

I think if you enable super sampling in the Vulkan renderer, it should allow you to increase the internal res while maintaining the proper integer scaling.

xchimp:
No, I'm running it at the native internal resolution, and the screen remains stretched no matter what I set the internal resolution to if using hardware rendering.

Where can I find that option to try it?

hizzlekizzle:
Ok, yeah, looks like it is indeed being weird. I'm able to reproduce the same behavior with no good explanation or solution.

xchimp:
That's good to know it's not only on my end then at least.

Silanda:
[...]

PR #551, opened 2019-10-18
Implement framebuffer height padding for GL renderer

ggdrt:
PR implements height padding for GL renderer. This fixes display height for games that don't draw to a full 240 lines framebuffer on original hardware and rely on what seems to be analog encoding to fill in the rest of the lines. Related to issue #525 though this PR only fixes it for the GL renderer and not the Vulkan one.

e.g. Wild Arms (draws to 232 height framebuffer)
**GL Renderer Before**: only framebuffer contents are displayed

**GL Renderer After**: properly letterboxed

Some lingering issues with the implementation: ~~loading save states can cause previous pixels to remain in the extra padded height area (might need to always clear the screen in rsx_gl_finalize_frame before calling bind_libretro_framebuffer, I'll test this out soon)~~ fixed and initial/last scanline options not working (they never worked with the GL renderer anyway; just a TODO).

### DIFF
--- a/mednafen/psx/gpu.cpp
+++ b/mednafen/psx/gpu.cpp
@@ -541,9 +541,17 @@ static void UpdateDisplayMode(void)
 
    uint16_t yres = GPU.VertEnd - GPU.VertStart;
 
+   bool is_pal_mode = false;
+   if((GPU.DisplayMode & DISP_PAL) == DISP_PAL)
+      is_pal_mode = true;
+
    // Both 2nd bit and 5th bit have to be enabled to use interlacing properly.
+   bool is_480i_mode = false;
    if((GPU.DisplayMode & (DISP_INTERLACED | DISP_VERT480)) == (DISP_INTERLACED | DISP_VERT480))
+   {
       yres *= 2;
+      is_480i_mode = true;
+   }
 
    unsigned pixelclock_divider;
 
@@ -588,7 +596,9 @@ static void UpdateDisplayMode(void)
          GPU.DisplayFB_XStart,
          GPU.DisplayFB_YStart,
          xres, yres,
-         depth_24bpp);
+         depth_24bpp,
+         is_pal_mode, 
+         is_480i_mode);
 }
 
 /* Forward decls */
@@ -1150,6 +1160,7 @@ void GPU_Write(const int32_t timestamp, uint32_t A, uint32_t V)
          case 0x07:
             GPU.VertStart = V & 0x3FF;
             GPU.VertEnd = (V >> 10) & 0x3FF;
+            rsx_intf_set_display_range(GPU.VertStart, GPU.VertEnd);
             break;
 
          case 0x08:
@@ -1853,6 +1864,8 @@ void GPU_RestoreStateP3(void)
    rsx_intf_load_image( 0,    0,
                         1024, 512,
                         GPU.vram, false, false);
+
+   rsx_intf_set_display_range(GPU.VertStart, GPU.VertEnd);
 
    UpdateDisplayMode();
 }
--- a/mednafen/psx/gpu.cpp
+++ b/mednafen/psx/gpu.cpp
@@ -1122,6 +1122,7 @@ void GPU_Write(const int32_t timestamp, uint32_t A, uint32_t V)
             GPU_SoftReset();
              rsx_intf_set_draw_area(GPU.ClipX0, GPU.ClipY0,
                                     GPU.ClipX1, GPU.ClipY1);
+             rsx_intf_set_display_range(GPU.VertStart, GPU.VertEnd); //0x10, 0x100 set by GPU_SoftReset()
              UpdateDisplayMode();
             break;
 
--- a/rsx/rsx_dump.cpp
+++ b/rsx/rsx_dump.cpp
@@ -130,7 +130,15 @@ void rsx_dump_set_draw_area(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1)
    write_u32(y1);
 }
 
-void rsx_dump_set_display_mode(uint16_t x, uint16_t y, uint16_t w, uint16_t h, bool depth_24bpp)
+void rsx_dump_set_display_range(uint16_t y1, uint16_t y2)
+{
+   if (!file)
+      return;
+   write_u32(y1);
+   write_u32(y2);
+}
+
+void rsx_dump_set_display_mode(uint16_t x, uint16_t y, uint16_t w, uint16_t h, bool depth_24bpp, bool is_pal, bool is_480i)
 {
    if (!file)
       return;
@@ -140,6 +148,8 @@ void rsx_dump_set_display_mode(uint16_t x, uint16_t y, uint16_t w, uint16_t h, b
    write_u32(w);
    write_u32(h);
    write_u32(depth_24bpp);
+   write_u32(is_pal);
+   write_u32(is_480i);
 }
 
 void rsx_dump_triangle(const struct rsx_dump_vertex *vertices, const struct rsx_render_state *state)
--- a/rsx/rsx_dump.h
+++ b/rsx/rsx_dump.h
@@ -16,7 +16,8 @@ void rsx_dump_finalize_frame(void);
 void rsx_dump_set_tex_window(uint8_t tww, uint8_t twh, uint8_t twx, uint8_t twy);
 void rsx_dump_set_draw_offset(int16_t x, int16_t y);
 void rsx_dump_set_draw_area(uint16_t x0, uint16_t y0, uint16_t x1, uint16_t y1);
-void rsx_dump_set_display_mode(uint16_t x, uint16_t y, uint16_t w, uint16_t h, bool depth_24bpp);
+void rsx_dump_set_display_range(uint16_t y1, uint16_t y2);
+void rsx_dump_set_display_mode(uint16_t x, uint16_t y, uint16_t w, uint16_t h, bool depth_24bpp, bool is_pal, bool is_480i);
 
 struct rsx_dump_vertex
 {
--- a/rsx/rsx_intf.cpp
+++ b/rsx/rsx_intf.cpp
@@ -224,6 +224,9 @@ struct DrawConfig
    int16_t  draw_offset[2];
    uint16_t draw_area_top_left[2];
    uint16_t draw_area_bot_right[2];
+   uint16_t display_area_yrange[2];
+   bool     is_pal;
+   bool     is_480i;
 };
 
 struct Texture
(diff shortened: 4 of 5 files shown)


ggdrt:
Nevermind, closing this for now.

Doesn't work properly for games that require analog encoding letterboxing at the top. Need to figure out how to access values from the GP1(07h) for a more robust implementation.

ggdrt:
Works for games requiring padding at the top now as well. Only implemented right now for 240p titles in the GL renderer.

Partial fix for #495 and #525 in the GL renderer.

**GL Renderer Before: only framebuffer contents are displayed**


**GL Renderer After: properly letterboxed**



The important part of the code I added is that the renderer needs to keep track of the Y display range that the game sets. If y1 = 16 and y2 = 256 then analog encoding begins displaying on the first displayable scanline and ends on (almost) the last one. Some games start at a larger y1 value, meaning that we need to add additional padding on top, or end at a smaller y2 value, meaning that we need to add padding on the bottom. 

I don't see any way to figure out how much padding is required with the variables that the GLrenderer had access to before this commit.

twinaphex:
Cool, do you feel confident in backporting this for the Vulkan renderer too. Or is that not necessary?

@rz5 Let me know your input as well.

rz5:
I think it's all good, ggdrt even created a new `rsx_intf` function so that other renderers can reuse the intercepted software GPU info.

I never thought about how to solve this problem so first I'd ask if the heuristic for non-standard fb heights is necessary:

```C
   if (_h < 240) 
   {
      top_pad     = (uint32_t) (renderer->config.display_area_yrange[0] - 16);
      bottom_pad  = (uint32_t) (240 - (_h + top_pad));
   }
```
Can't you unconditionally apply vertical padding based on what you intercepted from `GPU.VertStart` and `GPU.VertEnd`?

ggdrt:
I originally tried applying it unconditionally, but my code currently results in some strange behavior for non-standard 480-line content, so I restricted the problem set to 240-line content for now. I'll have to look into what's going on with the 480i stuff later.

@twinaphex The Vulkan renderer needs this implemented as well, but I don't know Vulkan enough to do it right now. The GL renderer is still missing a number of other features like width padding and top/bottom scanline cropping that I want to get implemented first. Those features also need to be added into the Vulkan renderer eventually.

ggdrt:
Think I have an idea of how to do this for non-standard 480-line content now, but the equation needs to be tweaked a bit for the 480 case. 

There's not really a foolproof way to differentiate between 240 and 480-line content inside the GL renderer that I like right now. `rsx_intf_set_display_mode` only sets numerical values so I'd have to change the case statement to something like `if (_h <= 240) {...} else if (_h <=480)` which is a little messier than querying a display mode status register.

I'm away at the moment but I'll get back to working on this some time before end of the week.

ggdrt:
@rz5 Should be working for all NTSC and PAL content in both 240p and 480i modes now. Some of the earlier commits had serious underflow bugs with PAL content since I wasn't differentiating between NTSC and PAL. 

The additions I made also sets the groundwork for allowing the Initial/Last Scanline cropping options to be implemented for the GL renderer in the future.

rz5:
It's looking better for sure.

Walk me through this if you don't mind. The issue is the missing lower black padding when using hw renderers.

Looking at the current state of this PR, the logic seems to be that when we're done drawing the GL frame by calling `bind_libretro_framebuffer()`, that is when you go in and pad the `display_resolution`'s height based on the values you got from `display_resolution_yrange`.

In the GL renderer's case, you add the missing padding by altering its viewport's lower left corner. I don't understand why offsetting the viewport lower left corner upwards would result in drawing the missing letterboxing...I guess you previously added more black lines than needed and you're trimming it as needed?

Instead of this approach, have you considered altering the `rsx_intf_set_display_mode` function instead? Or even changing `y` based on `GPU.VertStart`/`GPU.VertEnd` when it's called?
I suggest this because the PR will only cover the GL renderer. The hope is that if you correct the `y` value reported by `set_display_mode`, it would affect all current and future renderers using the rsx interface.

ggdrt:
So the problem isn't just lower black padding. It's that in general, the displayed output on both hardware renderers doesn't follow how the GPU+DAC on real hardware output framebuffers to analog video.

To the best of my knowledge, right now on master branch the GL renderer draws to some emulated VRAM (upside down due to different axes in PSX VRAM coordinate system vs GL coordinate system) and then slices out from the VRAM whichever one of the two PSX framebuffers it's supposed to display and draws that texture to the GL framebuffer provided by the frontend.

It draws that to a full 4:3 viewport, which is fine heightwise for NTSC 240p and 480i games that are well-conditioned: GP1(07h) display range y1=16, y2=256, which are the 240 (technically 243 I think, but that's not super important here) visible scanlines on an NTSC television. But this doesn't actually work properly for NTSC games with different values, like Threads of Fate and Wild Arms like we see in the pictures earlier in the thread. Threads of Fate uses some value like y1=30 and y2=238. That means of the 240 scanlines on an NTSC image, the GPU+DAC fills in 14 scanlines at the start, outputs the game's 208 height framebuffer, then fills in another 18 scanlines after that.

The current master branch behavior is even more inaccurate for PAL, where we should have 288 visible scanlines. Many games still draw to a 240 height PSX framebuffer and have their vertical display range values set to something like the middle 240 lines, relying on the GPU+DAC to fill in empty lines on the top and bottom. The GL renderer as described previously just takes the PSX framebuffer and stretches it all the way instead of doing any padding.

So in the modified code for `bind_libretro_framebuffer()`, we first calculate how many visible lines needed to be added in the final output on the top and bottom. We offset the viewport vertically by how much visible scanline padding is needed on the bottom, and also adjust the frontend resolution to a scaled multiple of native PSX output resolution, *not* a multiple of PSX framebuffer height (e.g. for Threads of Fate, we do a multiple of 240 instead of 208). This gets us our top padding for games that need it. 

Then back out in `rsx_gl_finalize_frame` after calling `bind_libretro_framebuffer`, the PSX framebuffer we want is displayed within the viewport configured in `bind_libretro_framebuffer`.

So going to your first question, it's not that we added more lines than needed, it's that we never had the lines needed in the first place.

As for the second question... adjusting rsx_intf_set_display_mode wouldn't really help here, and especially not adjusting the x and y values. The x and y values are coordinates that tell the GL renderer where the PSX framebuffer is located in VRAM. Adjusting that causes the wrong image to be displayed. You can't modify w or h either since those tell us how big the PSX framebuffer in VRAM is, and values that are too large cause garbage pixels to show up from stuff in VRAM that isn't meant to be displayed.

If there's a cleaner solution I'm all ears. Thanks for sticking with me through this.

ggdrt:
Making good progress on getting horizontal padding/cropping implemented for GL renderer.

rz5:
So you've identified 5 'standard' widths to which you can succesfully add bottom padding on HW renderers and optionally remove it if overscan cropping is on, right?

I hope you're making sense of it all. And be careful going down that rabbit hole of catching all edge-cases, if it's anything like PS2 games, looking at registers to try and deduce the final frame's dimensions may simply not work well due to the difference between displaying the PSX's framebuffer on an emulator's window and the analog encoding/decoding that happens between a PSX and a CRT screen.

Unless, of course, you're taking that analog processing into account.

ggdrt:
The five standard widths are for adding horizontal padding, right. In general you can hardcode the padding needed just by looking at the width mode that the game sets with the GP1(08h) command. There are a very small handful of games that you need to look at the GP1(06h) horizontal display range to figure out because they use a combination of very nonstandard horizontal display range values. I don't think it's worth testing the GP1(06h) values at the moment because 1. there are few games that have this issue and 2. the hardcoded values are for the most part good enough.

The vertical padding is calculated independently of any width padding that's needed -- these are actually two distinct features. For vertical padding we have to look at the GP1(07h) vertical display range to calculate. The only two vertical modes that we really have to consider are 240p and 480i.

Fortunately the PSX is far less complicated analog encoding/decoding-wise than the PS2.

@rz5 Thinking back on what you said earlier about said about altering `rsx_intf_set_display_mode()`, I do think there's at least some stuff that can be generalized across all renderers. For example, the padding calculations could be broken out into general utility functions that any of the renderers could call before they output the final image. Setting the viewport dimensions and such would probably still have to be renderer-specific though, unless there's any way at all to tell the libretro frontend to do something like "offset the provided GL/Vulkan framebuffer by X and Y amount" that I'm unaware of.

softtest03:
Would it be feasible to try and emulate the analog encoding/decoding and get the display dimensions "automatically" instead?

ggdrt:
Not exactly sure what you mean by getting the display dimensions automatically.

We're skipping over any actual analog NTSC scanline encoding by looking at what the PSX GPU is configured as and then displaying the framebuffer as how it would be shown on a 4:3 CRT. This isn't any sort of black magic -- a lot of this logic or similar logic is already being used by the software renderer or upstream Mednafen. The main work is just getting that working with the GL renderer. The only thing really new here is some rsx interface and GL-specific stuff that isn't in the software renderer.

softtest03:
It was my impression that you had to make adjustments for each resolution individually and account for a bunch of edge cases, and that emulating the analog encoding would avoid this problem. I was probably wrong though.

ggdrt:
The GPU+DAC on real hardware uses the width mode the GPU has been configured with to determine the pixel clock divider to output at. Every width mode the PSX has corresponds to a different pixel clock divider value. We're just emulating the behavior that would result from this analog encoding process, but not emulating the actual analog encoding itself.

There *are* some unusual cases yes, but they're somewhat rare. The GPU+DAC naturally handles this correctly, and it would be desirable to get this implemented but is not a personal priority for me right now. I just want to get the horizontal overscan cropping implemented for the general case, which covers the majority of games.

twinaphex:
Hi there, 

is this safe to be merged? @rz5 has this been put through its paces yet with a number of games?

ggdrt:
@twinaphex From my testing this should be safe to merge to get the height padding fixes into the GL renderer, but of course I can't be 100% confident without more user feedback. The last couple of comments in this thread have been about horizontal overscan cropping, which is a similar/related but separate feature.

After I have the rest of the related core options related to this implemented in the GL renderer, I can start factoring out any common code into utility functions that are reusable by the VK renderer.

twinaphex:
@ggdrt Alright, there seems to be one last merge conflict - probably as a result of @hizzlekizzle merging the other PR.

If you can go ahead and fix that one, we'll go ahead and merge this one. Thanks for your patience and thanks a lot for your contributions.

ggdrt:
@twinaphex Done -- do you mind squashing the commits in this PR for me? I'm away from my dev machine at the moment for the next day or so. Probably want to squash them since the intermediate commits have pretty bad bugs introduced that I ended up fixing with the later commits.

ggdrt:
@twinaphex Alright sorry for the delay, everything is good to go now. I rebased the PR branch and squashed the commits logically down to bug-free ones.

Issue #384, opened 2018-07-25
[OpenGL] Xenogears Heat Wave Effect broken 


Papermanzero:
Just checked and Xenogears’ ‘heat wave’ effect is still broken, guess it has nothing to do with mask bit.
For example, the sequence at the beginning of the game, when Lahan is attacked and set on fire, there is this very cool effect, like extreme heat haze. Beetle PSX HW renders the scene incorrectly, like so:

Reference: https://forums.libretro.com/t/beetle-psx-hw-mask-bit/17126

Screenshots:
![a3e57d95faf0c7e3798f67ff38b620afcfc0d989](https://user-images.githubusercontent.com/10263536/43199941-eaac0774-9013-11e8-9ee8-87ad82bdef98.jpg)
![e9a583419b9adbcc35efb2c2da74a4b53bf9d2ec](https://user-images.githubusercontent.com/10263536/43199946-ed7e8706-9013-11e8-9ea4-1ca9b261ceab.jpg)
![f4375ea9ef8d8cd46aa257238587a2f465138b43](https://user-images.githubusercontent.com/10263536/43199951-f03405ac-9013-11e8-8cdb-f4a664402d24.jpg)

CsoiretgeFsM:
Same problem here. Tested various settings: default, one by one, etc. Only seems to happen when "**Renderer**"  is set to _hardware_ in core options.

Setting "Renderer" to software fixes the bug but creates another by offsetting the PS bios intro to the left & up (black bar on the right side & bottom of white SCE logo). Everything after the bios intro isn't offest though and seems to display normally.

Soliaonline:
I have a monster of a compute running this and the part of the game where you have to go underwater is completely unplayable.  You cannot see anything.

If someone could look at this bug, I am happy to provide saves.

Issue #224, opened 2017-08-27
[Vulkan/OpenGL] Glitch in Chrono Cross effects


outoftime85:
I'm not sure if  this is related to the FF7/FF8 swirl reported in [Issue 199](https://github.com/libretro/beetle-psx-libretro/issues/199), or if this is the best place to file a bug report, but I hope this is helpful nonetheless.

I'm running Retroarch 1.6.7 with Beetle-PSX 0.9.44.1 09d2a02 (from the Stable PPA) on Ubuntu 17.04 with a Nvidia Geforce GTX 970 driver version 375.66.

There are several visual glitches that I've seen in Chrono Cross when running in either Vulkan or opengl. The most obvious is the effects when a battle starts, which I've captured here.

![Vulkan1](https://user-images.githubusercontent.com/31387895/29751181-7b9ab318-8b18-11e7-9b2c-2b707df8e635.png)
![Vulkan2](https://user-images.githubusercontent.com/31387895/29751182-7b9b8b3a-8b18-11e7-9441-f23186e56e6b.png)

![Opengl1](https://user-images.githubusercontent.com/31387895/29751180-7b9a6f84-8b18-11e7-859d-89b9b6680de4.png)
![Opengl2](https://user-images.githubusercontent.com/31387895/29751179-7b98c31e-8b18-11e7-9a41-a53d3f1c4010.png)
![opengl3](https://user-images.githubusercontent.com/31387895/29751178-7b984664-8b18-11e7-9550-50db9c5a202a.png)

The effect is normal in software.
![Software2](https://user-images.githubusercontent.com/31387895/29751184-8035f22a-8b18-11e7-8ba7-52d23d17e5ed.png)
![Software1](https://user-images.githubusercontent.com/31387895/29751183-8035b6f2-8b18-11e7-9777-a3acd0357ea0.png)

It may be hard to see with just screenshots, but I can work on figuring out screen recording if that would help. The main thing is there appear to be "windows" on the screen, boxes that I can see the effect moving within, but a lot of the rest of the screen in static. However [this video at 5:04](https://youtu.be/0Sgmr0lAzIw?t=5m4s) shows the effect working fine, apparently with Vulkan and PGXP (side note, the above occurs whether PGXP is enabled or disabled. I've also tried enabling/disabling software framebuffer).

[...]

PR #556, opened 2019-10-27
Make sure has_software_fb is set by GLRenderer_new

### LINKED ISSUE #246: [GL Renderer] BIOS screen glitches
Shortly after the Sony Computer Entertainment text appears, the text will disappear along with one of the logo's inner triangles and the jingle will be stuck in a loop, all of this in a span of 4 seconds.

The software and vulkan renderers do not have this bug. Is it perhaps a timing glitch? Maybe the value the GL renderer is using for the video clock doesn't match up with what the actual core uses?

### LINKED ISSUE #427: HW: Software framebuffer option doesn't work at start/restart in 32bpp mode, and is dithered when 32bpp mode is not dithered
FYI: In my experience, the Software framebuffer option (which I assume enables read-back effects?) doesn't work when a game is started. However, going into the options and setting any other option, makes it begin to work.

FWIW: I have a game that seems to save its picture, and use it as a background when you go into the game's menus. When I turn the game on, the background is always black. The only way to fix it is to go into Beetle's options and monkey with them. It's not enough to just open the Options screen, but changing anything seems to enable the background.

ggdrt:
Fixes #246
Fixes #427 

`GLRenderer_new` method was not setting `has_software_fb`, meaning that regardless of what the option was set to by the user, the GL renderer would always start with the software framebuffer off. The software framebuffer would have to be forced on by toggling any core option, causing `retro_refresh_variables` to be called, which has a check for `renderer_software_fb` in it. This PR fixes this issue by making sure `renderer_software_fb` is checked and `has_software_fb` is appropriately set by `GLRenderer_new`.

Also fixes the BIOS screen glitches since it turns out the "Sony Computer Entertainment" text requires software framebuffer to be on.

### DIFF
--- a/rsx/rsx_intf.cpp
+++ b/rsx/rsx_intf.cpp
@@ -1196,6 +1196,17 @@ static bool GlRenderer_new(GlRenderer *renderer, DrawConfig config)
    if (!renderer)
       return false;
 
+   var.key = BEETLE_OPT(renderer_software_fb);
+   if (environ_cb(RETRO_ENVIRONMENT_GET_VARIABLE, &var) && var.value)
+   {
+      if (!strcmp(var.value, "enabled"))
+         has_software_fb = true;
+      else
+         has_software_fb = false;
+   }
+   else
+      has_software_fb = true;
+
    get_variables(&upscaling, &display_vram);
 
    var.key = BEETLE_OPT(filter);


ggdrt:
Also turns out this is a partial fix for #224 and #384, but the behavior exhibited in those two threads is a compound mask bit + software framebuffer issue. This only solves half of the problem and there's already plenty of issue threads for mask bit emulation in the GL renderer so not sure if those two should be closed as well.

rz5:
Those other two issues yeah, not sure if your patch addresses them. Only by testing it, I guess.

However, great job on finding a fix for the first two bugs, for the longest of times I wanted to get that BIOS screen fixed and had no clue why it was buggy at the time.

I'll test out your changes with MGS and FF7.

ggdrt:
Yeah I tested Chrono Cross and Xenogears and was able to partially reduce the issues with forcing the software framebuffer on.

The issues are partially reintroduced on Vulkan when the software framebuffer is turned off, so they're at least partially tied to needing the software framebuffer to be on.

ggdrt:
Yeah I tested Chrono Cross and Xenogears and was able to partially reduce the issues with forcing the software framebuffer on.

The issues are partially reintroduced on Vulkan when the software framebuffer is turned off, so they're at least somewhat tied to needing the software framebuffer to be on.

rz5:
Yup, so without the patch and if I don't touch the sw fb option, FF7 battle swirls will be black, because `has_software_fb` was never changed from 'false' to 'true'. With the patch, we get battle swirls.

In hindsight, that static variable should, by default, be set to true. And perhaps the sw fb option was a mistake, I requested it hoping it would squeeze out more perf when using hardware renderers...

PR #557, opened 2019-10-28
Implement crop overscan and initial/last scanline options for GL renderer

ggdrt:
Two feature commits in this PR to cut down on the number of PRs I've been throwing out lately.

Implements horizontal overscan cropping feature and initial/last scanline feature. Unlike the software core, the initial/last scanline don't require restarting the core to take effect, which is pretty nice.

### DIFF
--- a/rsx/rsx_intf.cpp
+++ b/rsx/rsx_intf.cpp
@@ -330,6 +330,9 @@ struct GlRenderer {
     * the visible area */
    bool display_vram;
 
+   /* When true we perform no horizontal padding */
+   bool crop_overscan;
+
    int32_t initial_scanline;
    int32_t initial_scanline_pal;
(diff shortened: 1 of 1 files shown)


ggdrt:
There's still one small thing related to accuracy that I'm not entirely sure about. What happens if a game tries to display when the vertical range `y1 < 16` or `y2 > 256` ? The only example I can think of right now is when the GPU is reset, but in that case the display is off and nothing can be seen anyways. Right now, as implemented from a previous commit, we just bound check and clamp any rows existing outside of the displayable area to the edge of the displayable area, but that's probably not entirely accurate. 

In any case, with these two new commits in place the windowing stuff in the GL renderer is really close to being on par with the software renderer.

PR #559, opened 2019-11-01
Gl crop overscan accuracy additions

ggdrt:
PR does three things:

1. Hides overscan and scanline options currently not respected by each hardware renderer

2. Minor fix to properly handle games that try to display outside of the visible scanline region

3. Brings the GL renderer's **Crop Overscan: Off** to almost the same accuracy level as the Software Renderer's **Crop Overscan: Off** (barring differences related to the sw renderer updating on a scanline granularity versus the hw renderer copying the emulated vram fb as a texture)

Point number three was accomplished by modifying `UpdateDisplayMode()` and by extension `rsx_intf_set_display_mode()` to pass down an enum containing the width mode of the image to the renderer, analogous to how the GP1(08h) command sets a width mode of the image. This gives us a more reliable method for detecting the width mode of the image instead of using the unreliable `_w` variable as a heuristic. 

`rsx_intf_set_horizontal_display_range()` method was also added, which passes the horizontal start/end values down to the hw renderers and allows us to calculate how much horizontal padding is needed on each side at cycles-to-pixels granularity.

### DIFF
--- a/libretro.cpp
+++ b/libretro.cpp
@@ -3564,6 +3564,11 @@ bool retro_load_game(const struct retro_game_info *info)
          option_display.key = BEETLE_OPT(mdec_yuv);
          environ_cb(RETRO_ENVIRONMENT_SET_CORE_OPTIONS_DISPLAY, &option_display);
 
+         option_display.key = BEETLE_OPT(image_offset);
+         environ_cb(RETRO_ENVIRONMENT_SET_CORE_OPTIONS_DISPLAY, &option_display);
+         option_display.key = BEETLE_OPT(image_crop);
+         environ_cb(RETRO_ENVIRONMENT_SET_CORE_OPTIONS_DISPLAY, &option_display);
+
          break;
       }
       case RSX_VULKAN:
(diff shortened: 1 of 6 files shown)


ggdrt:
@rz5 Would appreciate your thoughts on this.

Working on these additions made me realize that currently `UpdateDisplayMode()` and `rsx_intf_set_display_mode()` are being used as an amalgam of the GP1(05h) and GP1(08h) commands. That is, `UpdateDisplayMode()` passes down info to `rsx_intf_set_display_mode()` that both GP1(05h) and GP1(08h) would pass to the GPU, but `UpdateDisplayMode()` is only being explicitly called whenever a GP1(08h) command is processed. 

No rsx interface function is being called when a GP1(05h) command is processed, meaning that we rely on the assumption that a GP1(08h) command will be called on every frame after any GP1(05h) commands. While this works for the most part, I'm not sure if this kind of behavior should be allowed. Implementing a new rsx interface method for GP1(05h) instead of relying on `UpdateDisplayMode()` being called by GP1(08h) might be the right way to make this GPU command behavior more explicit.

rz5:
Honestly, I would not be able to tell you what would be gained by intercepting GP1(05h) commands explicitly like you propose. 

It would help if you could highlight either a bug caused by it or a use-case where your crop/pad/overscan code benefits from it. Not that I would oppose it if you didn't, mind you.

Like you said, it seems to be working for the SW renderer and most of the times for the HW renderers.
And I agree, in theory it makes sense to intercept that command and pass it onto the HW renderers to keep the "Start of Display" coordinates fresh.

ggdrt:
The only thing I can think of at the moment for where we'd want to keep the FB start coordinates fresh is if we wanted to emulate the framebuffer swap timings more accurately like the software renderer does. This comes with its own issues such as introducing potential screen tearing which would be accurate but undesirable. In any case the hardware renderers don't process framebuffer outputting on a scanline granularity like real hardware and the software renderer do, so keeping the FB start coordinates constantly updated for the hardware renderers wouldn't affect anything by my guess.

It very likely wouldn't hurt to leave it be until the rsx interface ever needs a cleanup to disambiguate certain functionality, probably if there's ever another hardware renderer to be implemented.

PR #560, opened 2019-11-03
(SW renderer) Change base geometry on new display modes

rz5:
Currently, in the software renderer, base dimensions are always 320x240.
This behavior seems to not cause many issues. However, content base
dimensions do change and this commit makes the base dimensions track
those changes.


rz5:
This commit exposes a curious case in the US BIOS screen. While the PS logo is being rendered, display mode seems to be 640x472. Half a second later, when it is done rendering and the text appears, display mode switches to 640x**478**.

Perhaps the core geometry should only change if the PS1's display mode width is different?

ggdrt:
I think the height geometry should be set to only multiples of 240 (if 240p mode) or 480 (if 480i mode). 

If I'm remembering correctly, the reason why the display mode height seems to be smaller than 480 is because the BIOS, for whatever reason, doesn't actually set the vertical display range to the full 240 lines (doubled for 480i), it's doing something like [0x10, 0x99) instead of [0x10, 0x100) like most games would.

The w and h you're using here https://github.com/libretro/beetle-psx-libretro/pull/560/commits/8ac36e43afb2264966321ab2d6a2f06e7f94022a#diff-966d8fa253a8e8c699885d8edd7dbf02R4025 are calculated from the horizontal and vertical display range settings without accounting for any padding the GPU is going to fill in to create the full 240 or 480 line image.

negativeExponent:
i guess this is goodbye for window-mode users. forcing base_width/base_height to upscale size will force a bigger screen that its native 320x240 frame on integer scaling. mednafen cores used to separate the games geometry vs actual framebuffer geom (the ones sent to video_cb/retro_video_refesh_t) 

imho

ggdrt:
Actually now I'm wondering, what if instead you set the geometry info after `rsx_intf_finalize_frame()` is called in `retro_run()`? The video callbacks are done in finalize_frame and any last geometry changes are calculated by the GL renderer in that sequence of calls. Would calculating and setting any last geometry info after the video callbacks allow those changes to be applied in that particular `retro_run()` frame?

This seems the safest to me if it's possible, since by that point you know what that frame's display mode is and also if there's any overscan cropping or widescreen hacks being applied.

rz5:
@ggdrt Sure, I'll try using `is_480i` and `width_mode` instead and see how Castlevania reacts.

As for the observation for when to call `set_geometry`:  calling it after `video_cb` would probably be too late for the current frame - because it was submitted for rendering already - but I guess the new geometry would be seen on the next frame.

@negativeExponent You could still downscale it back to 320x240 with shaders, I guess? Right now the core never changes its base geometry despite submitting frames that are clearly smaller (256x240) and larger (640x478) than the reported base geometry.
Same thing for internal resolution upscaling, 320x240 would be thereported base geometry but in-game Castlevania would be submitting frames of size 512x480.

ggdrt:
This is quite a conundrum, not sure what the correct or desired behavior really is. If we set the geometry base width/height to the internally upscaled resolution, then the window scale setting in the RA frontend scales off of the upscaled geometry that we report. This means we can no longer do something like say, internally upscale by 8x and set window scale to 3x to get implicit supersampling down to 720p. 

Seems like there's a number of things that rely on the 320x240 reporting. The more I fiddle around to see what's going on, the more things start to act strangely.

hizzlekizzle:
> This means we can no longer do something like say, internally upscale by 8x and set window scale to 3x to get implicit supersampling down to 720p

AFAIK, this is how other hardware-rendered cores work, as well. That is, with windowed mode getting multiplied to enormity, though you can reduce the window size later. It's crummy behavior, but I'm pretty sure it's a frontend issue rather than the core's responsibility.

ggdrt:
Huh, looks that seems to be the case. I see that PPSSPP updates the geometry based on the internal resolution scale so that's at least one example in favor of this change.

rz5:
@ggdrt Pushed changes based on your observations about width/interlacing modes.

It was a quick and easy change because of your past changes to the `rsx_intf_set_display_mode` function. Something also occurred to me while doing this and reading the nocash Nocash PSX docs: this core is currently setting the max geometry as 700x576 but the biggest resolution mode seems to be 640x480.

I guess that 700x576 is like resolution of a virtual CRT monitor in which we can adjust the position of our 640x480-or-smaller frame?

hizzlekizzle:
I would assume that's for PAL interlaced 576i, though 700 is small for that. It's usually 720 or more.

rz5:
@hizzlekizzle The nocash docs suggest the PSX GPU is only capable of drawing 240 lines progressively or 480 interlaced. You'd know more about this than me but it seems that when a real PSX is in PAL video mode, some pixel clock trickery happens and those 480 interlaced lines are then "spread out" on a 578 canvas. Is this wrong?

In regards to the 700 value: it does seem to be wrong, I think it should be 768, as can be seen here: https://github.com/libretro-mirrors/mednafen-git/blob/master/src/psx/gpu.cpp#L141-L170

On my next commit to this branch, I'll adjust some values based on that block of code. PAL and NTSC seem to have different maximum heights, that will be part of the adjustment.

ggdrt:
@rz5 The 480 interlaced lines aren't spread out, they're just padded height-wise by the GPU analog encoder up to 288 per field for a full 576 per interlaced lines. This is something configured by the GP1(07h) display range command.

I'm not actually sure why 768 is used by mednafen, I can look into that in a while. The reason why 700 is being used as the max geometry width however is based on how many pixels are outputted in one active scanline region in 640 width mode. Mednafen approximates the number of GPU clock cycles in one active scanline region as 2800 cycles (it's ever so slightly more than that iirc, but 2800 gives a number that's divisible by all pixel clock values). In 640 width mode, the pixel clock divider is 4, meaning that the GPU outputs one pixel every 4 clock cycles -> 2800/4 = 700 pixels outputted in the active scanline region. This is also why when we do the overscan cropping for the software renderer in libretro.cpp, we crop off 30 pixels from the left side and another 30 from the right side to get only the 640 colored pixels, assuming that the display area is actually centered.

rz5:
I'm going to post some "video info" numbers and the actual width/height numbers submitted to the video callback while using the software renderer, all from a build of this branch while running Castlevania: SotN (USA) until in-game is reached.

I won't upscale, so max geometry will stay at 768x480 for the whole experiment. I'm also not using widescreen hacks, so the aspect ratio will stay at 4:3.

Type | base geom | frame geom
------|------------|--------------
BIOS | 640x480 | 700x480
FMVs | 320x240 | 350x240
Title screen | 512x240 | 560x240
Option screen | 368x240 | 400x240
In-game | 256x240 | 280x240

It does look like this branch is better at hinting what the height of its submitted frames will be relative to the master branch. However, because the frames are by default submitted with overscan, the reported base geometry is by default always wrong. I don't know the consequences of this.

And since we can detect things quite properly now, I'll bring back the subject of aspect ratios: we can easily it to anything more appropriate to the display mode we're currently on.
I've also noticed that forcing a 8:7 AR on PAL content gets it close to NTSC content at 4:3 AR (https://imgur.com/a/svYtdqG). Perhaps that's due to the vertical overscan added by the PAL mode...

twinaphex:
Looks like this PR needs to be rebased.

ggdrt:
@twinaphex I'm still in the process of separating out some stuff from the main goal of this PR. It's not a particularly easy rebase either because of the variety and amount of changes there are so it's still going to take some time. Depending on how that goes I may have to cherrypick the relevant changes to another branch or in worst case port the code over by hand.

ggdrt:
Going to close this for now. The way the core currently manages global state isn't exactly great and that needs to get sorted out for all of the builds first. There is also a lot of behavior contingent on header inclusion order that also needs to be fixed.

This is not a problem with the code in the PR specifically, just a problem with the way the repository code is organized that should be fixed before implementing these features.

PR #562, opened 2019-11-10
Implement display range code for Vulkan renderer

### LINKED ISSUE #346: [Vulkan] Alundra / Alundra 2 - Video Playback Offset
Settings

Renderer: Hardware (OpenGL)
Software Framebuffer: Off
Adaptive Smoothing: Off
Internal GPU Resolution: 4x
Texture filtering: bilinear
Internal Color depth: 32bpp
Wireframe Mode: Off
Display Full VRAM: Off
PGXP operation mode: Memory only
PGXP vertex cache: On
PGXP perspective correct texturing: On
Widescreen mode hack: Off
Frame duping: Off
CPU Frequency Scaling: 100%
GTE Overclock: Off
GPU rasterizer overclock: 1x
Skip Bios: Off
Dithering pattern: 1x
Display interal FPS: Off
Initial scanline: 0
Last scanline: 239
Initial scanline PAL: 0
Last scanline PAL: 287
Crop Overscan: On
Additional Cropping: Off
Offset Cropped Image: Off
DualShock Analog button toggle: Off
Port 1 Multitap: Off
Port 2 Multitap: Off
Gun Cursoer: Cross
Mouse Sensitivity: 100%
CD Access Method: sync
Memcard 0 method: libretro
Enable memory card 1: on
Shared memcards: off
Increase cd loading speed: 2x

GPU: Nvidia 760GTX

Description: Alundra and Alundra 2 (PAL version) have offsets during the playback of the intro and other video scenes. The video should be in the centre of the screen.
refernce: https://www.youtube.com/watch?v=gzJdlXPJaHs

Screenshot Alundra:
alundra-180624-205444

Screenshot Alundra 2:
alundra 2 - der beginn einer neuen legende-180624-205622

### LINKED ISSUE #495: Beetle PSX Hardware Rendering Stretches Video Which Breaks Interlacing

### LINKED ISSUE #525: Odd geometry behavior between SW & HW renderer on games with built in letterboxing

ggdrt:
Should be safe to merge, but I'll leave open to feedback for a bit.

There's two issues, one that's newly introduced and one that's a preexisting issue. 

The newly introduced issue is that with Crop Overscan **Off**, if a game tries to scroll itself outside of the maximum displayable horizontal area (e.g. when doing screen adjustment in Tomb Raider II) there's some minor flickering alongside the edge it's scrolling out of, possibly due to some rounding error logic that was already in the hardware renderer code. I'll look into this but it's not a priority as no game should really be displaying that close to the edge unless it's programmed poorly or it's trying to do screen adjustment for poorly calibrated CRTs.

The preexisting issue is that this implementation does not respect the horizontal display range values when Crop Overscan is **On**. This isn't a regression, just something that hasn't been fully implemented even for the GL renderer. Most games don't adjust the horizontal display range so the current behavior covers most cases but this is something to look into for future improvements.

Resolves #346 , resolves #495 , and resolves #525.

### DIFF
--- a/libretro.cpp
+++ b/libretro.cpp
@@ -3583,8 +3583,6 @@ bool retro_load_game(const struct retro_game_info *info)
          option_display.key = BEETLE_OPT(display_vram);
          environ_cb(RETRO_ENVIRONMENT_SET_CORE_OPTIONS_DISPLAY, &option_display);
 
-         option_display.key = BEETLE_OPT(crop_overscan);
-         environ_cb(RETRO_ENVIRONMENT_SET_CORE_OPTIONS_DISPLAY, &option_display);
          option_display.key = BEETLE_OPT(image_offset);
          environ_cb(RETRO_ENVIRONMENT_SET_CORE_OPTIONS_DISPLAY, &option_display);
          option_display.key = BEETLE_OPT(image_crop);
--- a/parallel-psx/main.cpp
+++ b/parallel-psx/main.cpp
@@ -443,8 +443,12 @@ static bool read_command(const CLIArguments &args, FILE *file, Device &device, R
 		auto w = read_u32(file);
 		auto h = read_u32(file);
 		auto depth_24bpp = read_u32(file);
+		auto is_pal = readu32(file);
+		auto is_480i = readu32(file);
+		auto width_mode = readu32(file);
 
-		renderer.set_display_mode({ x, y, w, h }, depth_24bpp != 0);
+		renderer.set_display_mode({ x, y, w, h }, depth_24bpp != 0, is_pal != 0, is_480i != 0,
+		                          static_cast<Renderer::WidthMode>(width_mode));
 		break;
 	}
 
--- a/parallel-psx/renderer/renderer.cpp
+++ b/parallel-psx/renderer/renderer.cpp
@@ -675,6 +675,49 @@ void Renderer::mipmap_framebuffer()
 	}
 }
 
+Renderer::DisplayRect Renderer::compute_display_rect()
+{
+	unsigned display_width;
+	unsigned display_height = (render_state.is_pal ? 288 : 240) * (render_state.is_480i ? 2 : 1);
+
+	int left_offset = 0;
+	unsigned clock_div = 10;
+	switch (render_state.width_mode)
+	{
+	case WidthMode::WIDTH_MODE_256:
+		clock_div = 10;
+		break;
+	case WidthMode::WIDTH_MODE_320:
+		clock_div = 8;
+		break;
+	case WidthMode::WIDTH_MODE_512:
+		clock_div = 5;
+		break;
+	case WidthMode::WIDTH_MODE_640:
+		clock_div = 4;
+		break;
+	case WidthMode::WIDTH_MODE_368:
+		clock_div = 7;
+		break;
+	}
+
+	if (render_state.crop_overscan)
+	{
+		// Horizontal crop amount is currently hardcoded. Future improvement could allow adjusting this.
+		display_width = 2560/clock_div;
+		left_offset = (render_state.horiz_start - 608) / (int) clock_div;
+	}
+	else
+	{
+		display_width = 2800/clock_div;
+		left_offset = (render_state.horiz_start - 488) / (int) clock_div;
+	}
+
+	int upper_offset = (render_state.vert_start - (render_state.is_pal ? 20 : 16)) * (render_state.is_480i ? 2 : 1);
+
+	return DisplayRect(left_offset, upper_offset, display_width, display_height);
+}
+
 ImageHandle Renderer::scanout_to_texture()
 {
 	atlas.flush_render_pass();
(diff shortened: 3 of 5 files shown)


rz5:
I just tested beetle-psx, with and without this PR, with Chrono Chross (USA) Disc 1, making early in-game in a battle.
The "Meteor Shower +1" causes a little screen shake effect with the software renderer and that's accurate behavior.

Using the Vulkan renderer without this PR, I get no letterbox/pillarboxes and the screen does not shake.
With the PR, the Vulkan renderer, I get the letterbox/pillarboxes and the screen shakes. "Crop overscan" only affects the pillarboxes though.

Looking good to me.

hizzlekizzle:
I asked our resident Vulkan expert about it and he had these suggestions:

> https://github.com/libretro/beetle-psx-libretro/pull/562/files#diff-5dfb22247554410c870c7ddc6a7069aeR960 needs a barrier before it
> otherwise, it just seems like a crop of some sort
> not sure why all the sizing logic needs to be in there
> it could just be a "rect for scanout" parameter or something from the outside
> there's also a redundant copy
> which is a large bandwidth drain at high resolutions

ggdrt:
@hizzlekizzle These changes should be more efficient. Wondering how this looks now.

So the basic changes here are that we increase the render target image size to be the resolution that we want to output at. We then create a barrier and clear the render target image (to prevent any unwanted output in the pillarboxes or letterboxes from reusing the previous scanout in the original code, that could occur e.g. when loading save states). The original code then creates render pass information from that image and inserts another barrier before beginning a render pass. In the render pass, I set the viewport so that we render only to the particular area of the image that we want to. Much cleaner looking fix than what I initially had.

There's probably a way to do a clear within the render pass but I wasn't able to figure out in my free time today how to do it with the existing code. There's also some display range code that can be factored out into the RSX functions that I left in there just for testing purposes. I'll clean that up later, just want to check if this looks better.

ggdrt:
Alright this seems pretty solid now, and a lot cleaner to boot -- inside the render pass instance, we clear the color attachment using the command buffer's `clear_quad` method. Then we adjust the viewport and we get the output that we should.

If this is good then I'll start factoring out the GL/Vulkan common code and clean up the commit history as well.

ggdrt:
Alright this seems pretty solid now, and a lot cleaner to boot -- inside the render pass instance, we clear the color attachment using the command buffer's `clear_quad` method. Then we adjust the viewport and get the output that we should.

If this is good then I'll start factoring out the GL/Vulkan common code and clean up the commit history as well.

ggdrt:
Alright this seems pretty solid now, and a lot cleaner to boot -- inside the render pass instance, we clear the color attachment using the command buffer's `clear_quad` method. Then we adjust the viewport and get the output that we should.

If this is good then I'll start factoring out the GL/Vulkan common code and clean up the commit history as well.

ggdrt:
Squashed and merged with optimized changes, tested and verified by rz5.

Issue #416, opened 2018-08-30
User should be informed when software rendering is used as a fallback (and why)


brianblakely:
If the user is not using either the OpenGL or Vulkan video driver, then Beetle PSX HW automatically falls back to software rendering.  A user unaware of this behavior may believe that the core is running in hardware mode, and that their device is simply under-performing.

The user should be notified that this fallback happened, and how to fix it.  Something similar is done when slow CD reads are detected: a notification tells the user to set their CD read mode to "precache".

ggdrt:
Added in 748d4f8 with some more messages added in 62b6333.

PR #573, opened 2019-12-04
Fix GET_PREFERRED_HW_RENDER handling and fallback

ggdrt:
This issue has a bit of an odd history. For some period of time before October, the core was forcing Vulkan for the hardware renderer even if the frontend driver was set to OpenGL. After Rinnegatamante's changes in October, the core was still forcing Vulkan, but now *only if* the frontend requested context wasn't OpenGL. That meant setting the D3D9, D3D10, etc. drivers in RetroArch would still cause the core to force Vulkan if possible when the actual desired behavior is for the core to fall back to the software renderer. Additionally, if RETRO_ENVIRONMENT_GET_PREFERRED_HW_RENDER was not supported by the frontend, the core would force Vulkan, and only attempt Vulkan, no matter what.

This PR cleans this up by doing a number of things.

1. If the frontend supports RETRO_ENVIRONMENT_GET_PREFERRED_HW_RENDER, the core will attempt to initialize the requested hardware renderer. If it can not because the hardware renderer for that video driver either does not exist or because of some other issue, the core will fall back to the software renderer.
2. If the frontend does not support RETRO_ENVIRONMENT_GET_PREFERRED_HW_RENDER, the core has no information as to which renderer it should create and so it will go down the list one by one until it finds a renderer that it is able to create. If it can not find any, the core will fall back to the software renderer.
3. If the core needs to fall back to the software renderer, it will display a message informing the user of why it is doing so (#416). The message displaying could use some work, suggestions on how to get this to look nice would be much appreciated.

### DIFF
--- a/rsx/rsx_intf.cpp
+++ b/rsx/rsx_intf.cpp
@@ -3721,17 +3721,26 @@ bool rsx_intf_open(bool is_pal, bool force_software)
          software_selected = true;
    }
    else
+   {
       /* If 'BEETLE_OPT(renderer)' option is not found, then
        * we are running in software mode */
       software_selected = true;
+   }
 
    if (!software_selected)
    {
-      unsigned preferred; // This will be set to a const value if GET_PREFERRED_HW_RENDER is not supported by frontend
-      if (!environ_cb(RETRO_ENVIRONMENT_GET_PREFERRED_HW_RENDER, &preferred)) preferred = 0xFFFFFFFF;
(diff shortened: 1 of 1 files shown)


ggdrt:
As an aside it seems possible to implement a restart-only core option that lets users force a hardware renderer, e.g. having selections like Auto/OpenGL/Vulkan. It might help clear up the confusion some users have of how to select which hardware renderer to use. Not the goal of this PR, just a future possibility.

hizzlekizzle:
Excellent additions, and I think a "force renderer (restart)" Auto|OpenGL|Vulkan|Software option is a great idea.

Issue #501, opened 2019-05-08
[Beetle PSX HW] Lightgun API is broken


Phiben:
Since the release of the Lightgun API  #271 by @hiddenasbestos  back in 2017 it never worked in Beetle PSX HW.

It kind of works in the non-HW version of the core but only at native resolution. If the resolution is higher than native, the gun alignment is completely off.

In Beetle PSX HW (OpenGL and Vulkan) it simply doesn't work at all. Neither at native resolution nor at upscaled resolutions.

chasburger:
Yeah I've run into this as well. Lightgun functionality in Bettle PSX HW when using the OpenGL or Vulkan driver is non-existent. The trigger and button inputs are sent, but the cursor is not drawn..

I can confirm that using higher than native internal resolution in the non-HW version messes with the alignment as well.

I will say that the lightgun API functions in Beetle PSX HW if you use DX11 or 12 though. Cursor is drawn and everything works as it should. So I just use that as a workaround for now. Slightly irritating to have to load another video driver and restart the RetroArch frontend, but it's not the end of the world.

rz5:
@chasburger Beetle PSX HW only has hardware renderers for OpenGL and Vulkan. If you're using RetroArch with a video driver different from gl, glcore and vulkan, that's solely software rendering.

ggdrt:
Diagnosis of the alignment issues: https://github.com/libretro/beetle-psx-libretro/issues/471#issuecomment-558778374

As for the lightgun not showing cursor for hardware renderers, that's probably going to require an RSX interface hook in `mednafen/psx/frontio.cpp` for either the `crosshair_plot()` or `DrawCrosshairs()` function. These draw functions also need to be adjusted to scale the cursor's size up proportionally for the software renderer at higher internal resolutions.

Darknior:
Hi everyone, some news about this problem ? Some solution to solve it ?
Thanks

Tovarichtch:
I've tested in software rendering and indeed, it works. Hardware rendering is the problem here, whatever the API used. It would be great to fix this, since it's the best core with all the light guns required to play all gun games (including Area 51 which is not possible to play without the Justifier, and it has no GunCon conversion patch).

SeongGino:
Can confirm that this still seems to be broken with hardware rendering specifically, as it works totally fine in software mode (with all the performance caveats that it brings).

[...]

Issue #471, opened 2018-12-16
[Bug] Gun crosshair works correctly only in native resolution


voradori:
###  Description

Crosshair is behaving wrong on X-axis with internal resolution >1

### Expected behavior

Crosshair correctly follows mouse cursor and real ingame aim mark on any GPU render scale

### Actual behavior

a)
Crosshair's position on X axis is divided by N when using N image scaling. 
E.g. when scale = 2, and mouse is in the center, crosshair is between cursor\mark and left screen edge

Y axis works as expected.
It's just crosshair X position draw bug - real aim mark is where it should be. 

![beetle psx](https://user-images.githubusercontent.com/5952315/50057606-fab1b200-017d-11e9-8c5d-4978901658c2.png)

b)
There is also a very slight difference between crosshair and aim mark (and mouse cursor when windowed) even on native internal resolution -  the closer crosshair to left\right edge is, the bigger is the difference. But it's very minor thing - the biggest difference is about 0.5 cm on a standard fullhd 24" screen

![beetle psx 2](https://user-images.githubusercontent.com/5952315/50057642-4d8b6980-017e-11e9-98ac-6bfcfba14d9b.png)
![beetle psx 3](https://user-images.githubusercontent.com/5952315/50057643-4d8b6980-017e-11e9-9027-c9108c073eab.png)

### Steps to reproduce the bug

1. Set input device 1 to Guncon
2. Run Time Crisis to calibration screen
3. Shoot at the aim to spawn red aim mark
a)  4. Set internal gpu resolution to any value > 1
b)  4. Let Internal gpu resolution at 1
5. Move crosshair to Left and Right edges (try in both windowed and fullscreen)
### Version/Commit

0.9.44.1 beeee12 SW\HW core 
(There's no working crosshair and even shots under mouse cursor on HW core for gl\vulcan renderer, even in SW mode. DX renderers works with same problems as SW core)

### Core options
Internal GPU Resolution = 2
Skip Bios = True
Widescreen hack = true

### Environment information

- OS: Win10 Pro x64 1809
- Retroarch 1.7.5

Jj0YzL5nvJ:
Is the widescreen hack. That option only works correctly on some 3D games. The "aim" is a 2D element.

voradori:
don't judge with no testing. 
everything is the same for WS hack off

Jj0YzL5nvJ:
This recall me a very old bug on Linux with the first generation of Intel HD Graphics...
[...]

PR #575, opened 2019-12-09
Lightgun Fixes

ggdrt:
An attempt to solve #471 and #501.

**2x IR scaling before** (note the halved width on cursor, cursor only displaying every other line, improper horizontal positioning)

**2x IR scaling after**

**Crop Overscan before** (improper horizontal positioning)

**Crop Overscan after**

**PAL before** (improper vertical positioning)

**PAL after**
capture-pal-fix

### Remaining Issues
1. Drawing cursor is still yet to be implemented for hardware renderers. I think lightgun games are only really going to work for the hardware renderers if Software Framebuffer is enabled unless we hook in some hard GPU readbacks like what maister was doing a while ago. Getting the lightgun to work with the hardware renderers in the first place is going to require some re-architecting of how we track global state in the core.
2. Scaling formula not modified for initial/last scanline core options. I don't really plan on implementing that until we have a better scaling formula.
3. Haven't tested the Justifier color scaling fix but it should work properly.

### DIFF
--- a/Makefile.common
+++ b/Makefile.common
@@ -170,7 +170,8 @@ ifneq ($(HAVE_GRIFFIN), 1)
       SOURCES_CXX += $(CORE_EMU_DIR)/decomp.cpp
    endif
 
-   SOURCES_C += $(CORE_DIR)/libretro_cbs.c
+   SOURCES_C += $(CORE_DIR)/libretro_cbs.c \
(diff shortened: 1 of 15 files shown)


ggdrt:
Cleaned up some extra comments I forgot to remove. Going to merge this in later today if there aren't any objections.

Issue #588, opened 2020-01-20
(Dynarec) Known bugs/issues


twinaphex:
Tested on Windows (x64) -

# Parasite Eve 2 (USA) (Disc 1)
Won't continue from start until you set CPU type to Beetle Interpreter

# Brave Fencer Musashi (USA)
Hangs at 'New Game' - first world screen

# Gex - Enter The Gecko (USA)

Crashes after PSX copyright screen

#Time Commando (USA)

Corrupted background graphics. Background graphics in this game are streaming FMV with polygon models on top.

# Street Fighter Ex Plus Alpha (USA)

If you lose a fight and then select Continue, it will hang

simias:
You're probably already aware of this but there are a bunch of comments regarding Viewpoint in mednafen/beetle's code (just grepping 'Viewpoint' will show them). Apparently it's shoddily coded and very timing-sensitive.

rz5:
Yeah, it's because of the comments in the DMA code mentioning it that I specifically tested it. Running it and using lightrec's dynarec **and its interpreter (!)** will cause a crash but the same doesn't happen on pcercuei pcsx-rearmed fork, which might be indicative of lightrec integration on beetle-psx.

DonelBueno:
PGXP doesn't seem to work with any of the lightrec modes. All of its benefits are lost when you enable lighrec.

tabnk:
Android Build
Tested on Galaxy S7.

OpenGL:
Unable to open or find hardware renderer. Auto switch to software mode

kikeminchas:
Crash Bandicoot 1 and 3 (ARM64, Nvidia SHIELD TV, Vulkan, Android 9)

Graphics are missing triangles / Large amount of triangles flickering constantly (Interpreter working fine)

Dynarec options as default / Mode: Cycle Timing check or Max Performance (happens in both)

simias:
@DonelBueno I'm fairly sure that it's simply not implemented in the Dynarec. You need CPU support for that stuff.

twinaphex:
Hi there @iCatButler ,

could you perhaps contact us and meet us on Discord? The Beetle PSX dynarec guys are stuck on how they'd be able to add PGXP support to the dynarec, and they'd really appreciate somebody like you being able to brainstorm with them on how they go around fixing this issue. The persons in question are zcrc and @ZachCook .

In case you're interested, please contact Autechre on Libretro discord and I'll add you to the chan.

[...]

PR #590, opened 2020-01-21
Lightrec Dynarec

ZachCook:
Disabled by default, compile with HAVE_LIGHTREC=1 to enable, and select "Max Performance" for CPU dynarec option (beetle_psx_cpu_dynarec = "execute" in .cfg)

Bugs should be reported in #588

See the .gitrepo files for the source of the new folders in deps, they are separate projects being added as git-subrepos for ease of building


ZachCook:
I will squash before this should be merged

PR #605, opened 2020-02-03
Add disk control interface v1 support

jdgleaver:
This PR adds full support for libretro disk control interface v1. This includes the following enhancements:

- Disk names are now shown when swapping disks via the `Current Disk Index` drop down list

- When loading multi-disk content (M3U playlists or PBP files), the last used disk index is now automatically saved/restored

### DIFF
--- a/libretro-common/include/libretro.h
+++ b/libretro-common/include/libretro.h
@@ -1,4 +1,4 @@
-/* Copyright (C) 2010-2018 The RetroArch team
+/* Copyright (C) 2010-2020 The RetroArch team
  *
  * ---------------------------------------------------------------------------------------
  * The following license statement only applies to this libretro API header (libretro.h).
@@ -1117,7 +1117,7 @@ enum retro_mod
                                             * This may be still be done regardless of the core options
                                             * interface version.
                                             *
-                                            * If version is 1 however, core options may instead be set by
+                                            * If version is >= 1 however, core options may instead be set by
                                             * passing an array of retro_core_option_definition structs to
                                             * RETRO_ENVIRONMENT_SET_CORE_OPTIONS, or a 2D array of
                                             * retro_core_option_definition structs to RETRO_ENVIRONMENT_SET_CORE_OPTIONS_INTL.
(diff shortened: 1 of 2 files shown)


simias:
I like this a lot! Is there a long term plan to move all the CD image handling code to the frontend? It would be nice if the generic CD handling (image formats, CHD decompression, physical CD drive handling etc...) could be factored away from the individual cores. Of course that would add a pretty large burden on the frontend but conceptually I think it'd make sense for it to be there.

jdgleaver:
@simias I don't think anyone's planning that at the moment, but you make a good point there  :)

The issue at present is two fold:

- We are constrained by the current API. We can add small extensions (like this disk control v1 add-on), but we have to be very, very careful about maintaining backwards compatibility, and compatibility with other frontends (it's easy to forget that there's a whole ecosystem making use of the libretro API, not just RetroArch!)

- The frontend - by design - does not know any specifics about any particular core. It can pass content either as a data buffer or a file path, and that's fine - but if you make the frontend responsible for handling whatever esoteric file formats a core may require, you really start going down the rabbit hole...

On the API side, the idea is to work on a revamped 2.0 release in a year or two - this will most likely change many things compatibility-wise, and should allow many current limitations to be lifted. So there's certainly the potential to rethink content handling here.

Regarding data formats, it's still a bad idea for the frontend to directly process arbitrary content - but if it's possible to do so, then it may well make sense to handle additional ubiquitous compression formats (like CHD) and things as standard as CD/DVD images. It's certainly something to think about!

Kupo91:
> When loading multi-disk content (M3U playlists or PBP files), the last used disk index is now automatically saved/restored

Unfortunatelly this doesnt work for me. Tested with Grandia as m3u playlist. Swapping works but it doesnt save the last used disk index. Or it doesnt restore it. Everytime I start the game I start with disk 1.
Here is a log file. Although I didnt find any errors in there.
[retroarch__2020_02_26__00_30_35.log](https://github.com/libretro/beetle-psx-libretro/files/4252677/retroarch__2020_02_26__00_30_35.log)

jdgleaver:
@Kupo91 You are using an old version of RetroArch that doesn't have the required disk control API extension.

Please update to the latest nightly.

PR #608, opened 2020-02-06
Italian core options

ggdrt:
Tested and should be safe to merge.

@jdgleaver Wanted to ask you something about core options -- so in libretro.h, we have a guarantee that any `retro_core_option_definition` structs not found in `local` will be read from `us` instead.
https://github.com/libretro/beetle-psx-libretro/blob/58e4fae9099a29d0ea71e7544c942978d35b181f/libretro-common/include/libretro.h#L1217-L1222

However, I didn't find anything in that file for guarantees when the `values` array in a specific `retro_core_option_definition` struct does not match that of its localization.

I noticed that when for example we have the following `values` array in a `retro_core_option_definition` struct in `us`
```
{
   { "1", NULL },
   { "2", NULL },
   { "3", "US" },
   { NULL, NULL },
}
```
and the following `values` array in a corresponding `retro_core_option_definition` struct in `local`
```
{
   { "3", "LOCAL" },
   { "4", NULL },
   { "5", NULL },
   { NULL, NULL },
}
```
the final displayed result in RetroArch has the range of the first `values` array but uses the localization from the latter `values` array, almost as if it's parsing `us` to get the valid range and applying any localization if it exists while ignoring every value in the localization that isn't in the US English core option definition. Is this behavior that can be expected or guaranteed? If so then that would make the localizations a bit easier to maintain since we would only need to include the bare minimum of values that need to be translated, with any changes to the default language core options propagating through to all localizations.


jdgleaver:
@ggdrt Yes, that's exactly how I designed it - sorry for not making it clear in the comments!

In all cases, the machine-readable `key` and `value` elements are taken from `option_defs_us`. The individual localised structs are then searched for these elements - whenever they are found, the translated human-readable `description` and `value_label` strings are used as an override.

So it is absolutely the case that you only need to translate the bare minimum, and all missing entries will propagate through from `option_defs_us`. This is guaranteed  :)

Issue #563, opened 2019-11-15
[Beetle PSX HW] Frame Duping core option not implemented for HW renderers


ggdrt:
Frame duping core option is currently not implemented for both GL and Vulkan hardware renderers and so toggling the core option has no effect on them. I have a commit that hides this option for those renderers that I'll PR alongside some other commits.

This link seems to imply this functionality is possible for hardware renderers: https://github.com/libretro/RetroArch/blob/666ebb0ae10db5962f6c6e99d9495a2e9deea093/libretro-common/include/libretro.h#L2150-L2153

I'll implement and test this sometime in the future, but hiding the option should suffice for now. Leaving this issue here for reference.

ggdrt:
Implemented for Vulkan, needs implementation for GL now.

Issue #509, opened 2019-06-02
[Beetle PSX HW] Correct Refresh Rates (game speed)


Dogway:
## Description

The hard-coded console refresh rates are incorrect for both NTSC and PAL regions, leading to incorrect game speed with libretro's default time-stretch settings.

### Expected behavior

hunterk merged a commit under a [fork ](https://github.com/hizzlekizzle/beetle-psx-libretro/commit/d82a3cdc14185f8aac9bf71e85c838a023bab8ae).
I did some calculations on this [post](https://forums.libretro.com/t/retroarch-crt-switchres-multiple-monitor-issue/17943/30), but I believe hunterk's PAL values are more correct (accounted for correct PAL width)

Papermanzero:
Can those refresh rates be used to speed up the PAL versions of certain games

Dogway:
You can already speed up PAL games using RA's timing_skew option, the problem is the core is reporting a wrong frame rate for both NTSC and PAL so you won't play either region games at their original speed (or extrapolated speed in case of PAL unoptimized games) if you wanted, unless you want to do some hackish maths to compensate.

Dogway:
Does the maintainer accept bounties?

twinaphex:
Of course.

hizzlekizzle:
Ok, the main reason I hadn't PRed this myself is that it wasn't automatically switching for interlaced content, but you're saying that's not desirable anyway, right? So how should the option be named/worded? Which speeds do we need to expose?

Dogway:
For CRT I guess it should switch.

For now I would set them to the progressive rates that matches with most of the content and open a ticket for the switch on interlaced content, because among other things I'm not sure how the switch will affect those with modelines. For instance if it boots in interlaced refresh rate the modeline won't be triggered. Not sure if libretro can know what kind of display (EDID?) is being used, or at worst add a core option as "Allow Interlaced Rate" or else because there might be someone who feeds native resolution to an LCD as well.

ggdrt:
[...]

PR #612, opened 2020-02-13
Add frame duping support to RSX Vulkan

ggdrt:
Provides some minor performance increase for titles that don't swap PSX VRAM framebuffers at a full ~50 or ~60 Hz refresh rate.

A consequence of enabling this option is that some FMV and BIOS video bugs (occasional garbled 24-bit/15-bit conversion issue very first frame of FMVs, PS BIOS logo teleporting into the bottom right corner at the end of playing the BIOS) are alleviated, but I think this option really only addresses the symptoms and not the cause, so those issues can't be closed on the tracker yet.

I tried implementing this in the GL renderer as well but was tripping over graphical issues left and right, so any help there would be greatly appreciated.

### DIFF
--- a/libretro.cpp
+++ b/libretro.cpp
@@ -4127,9 +4127,6 @@ bool retro_load_game(const struct retro_game_info *info)
          option_display.key = BEETLE_OPT(image_crop);
          environ_cb(RETRO_ENVIRONMENT_SET_CORE_OPTIONS_DISPLAY, &option_display);
 
-         option_display.key = BEETLE_OPT(frame_duping);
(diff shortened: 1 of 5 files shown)


ggdrt:
PR addresses Vulkan portion of #563, also sneaked in a commit that standardizes the timings by default to the non-interlaced rates for both NTSC and PAL (partially addresses #509).

ggdrt:
> A consequence of enabling this option is that some FMV and BIOS video bugs (occasional garbled 24-bit/15-bit conversion issue very first frame of FMVs, PS BIOS logo teleporting into the bottom right corner at the end of playing the BIOS) are alleviated, but I think this option really only addresses the symptoms and not the cause, so those issues can't be closed on the tracker yet.

This particular issue has been resolved by 807bc73

PR #613, opened 2020-02-15
Add support for forcing hw rendering backend

ggdrt:
Feature discussed back in #573 that's also been requested by users many times on various channels.

The logic still works the same way as before for the `Hardware` option value, which is renamed `Hardware (Auto)`. If the frontend's preferred driver doesn't have a corresponding hardware renderer, we fall back to the software renderer and notify the user, and if the frontend doesn't have a preferred driver, we go down the list trying to initialize a hardware renderer instance until we get one.

The two new option values, `Hardware (Vulkan)` and `Hardware (OpenGL)` work with pretty much the same logic as the frontend having a preferred driver. We try to open the user's selected hardware renderer, and if we're unable to do so we simply notify the user and fall back to the software renderer. These two new values take precedence over the frontend's preferred driver and will cause that preferred driver setting to be ignored.

@hizzlekizzle @rz5 Would be appreciated if you guys have time to test this in case I missed something.

### DIFF
--- a/libretro.cpp
+++ b/libretro.cpp
@@ -3344,7 +3344,7 @@ static void check_variables(bool startup)
       bool hw_renderer = false;
       if (environ_cb(RETRO_ENVIRONMENT_GET_VARIABLE, &var) && var.value)
       {
-         if (strcmp(var.value, "hardware") == 0)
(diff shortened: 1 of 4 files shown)


ggdrt:
Did another round of testing and it's not showing any obvious issues so I'm going to go ahead and merge this in.

Issue #67, opened 2016-06-14
Feature request: allowing users to manually select Region / Video Mode (PAL, NTSC, AUTO)


Ryunam:
This is to kindly inquire about the possibility of exposing an option to select the Region / Video Mode in the Core Options of beetle-psx-libretro, thus allowing the user to either force a specific format (PAL / 50hz or NTSC / 60hz) or leave it up to the automatic region detection normally performed by the emulator.

The reason behind this request lies in being able to circumvent the technical limitations that affected the PAL-exclusive localizations of several classics.
Back in the day, PAL PS1 games were released on the market in a crippled form that capped them at 50hz, making the overall framerate 17,5% slower if compared to their respective NTSC counterparts. Furthermore, because of the intrinsic difference between the two formats and the different number of lines displayed, PAL versions were often plagued with black borders.

While it appears that the latter effect is mitigated in emulation, the former is still firmly in place. 
That is because, as far as I understand, Mednafen PSX faithfully reads the regional code of each game and then proceeds to run it with the corresponding Region / Video Mode, but this makes it so that some German, French, Italian and Spanish translations are inevitably hindered by the slower framerate. 
Being able to choose between modes would allow a wider array of options, while keeping the goal of accuracy.

Thanks in advance!

simias:
That sounds trivial enough to implement, a "force PAL/NTSC" option in the core. I'll give it a try.

simias:
On 2nd thought just changing the detected video format won't be enough, you'd also have to hijack the game to prevent it from setting the PAL bit in the GPU DisplayMode (all PSX GPUs can output both PAL and NTSC, there's a register to configure this behaviour).

Ryunam:
I see. In fact one thing I want to highlight here is that I had attempted to patch my PAL localized games to NTSC through some tools available around (which I assume have the effect of changing the GPU DisplayMode you referenced), but since Mednafen automatically detects the video format in an automatic way they would play in their usual crippled speed (50hz).

I wonder if in the meantime the option you have mentioned in your first post could be implemented anyway, so as to test how it behaves with NTSC-patched images. 

Or, another idea that comes to my mind, would it be possible to make it so that when you activate a forced Region / Video mode from the Core options, it would also impose the same configuration for the GPU bit?

needo2:
> PAL versions were often plagued with black borders

[...]

PR #632, opened 2020-03-24
Added "fast PAL" hack to allow PAL games to play at NTSC framerates

simias:
This is an implementation of the feature request #67 

It adds a core option "pal_video_override" that can be used to hack the video timing of PAL games in order to make them run at 60fps instead of 50. The resolution remains the same. NTSC games will ignore that setting entirely so it shouldn't have any impact for them.

As discussed in the issue many (but not all) PAL ports of NTSC games also have black bars at the top and bottom of the screen. This patch will not remove them. An other patch will be necessary to optionally crop the video output to remove those bars.

### DIFF
--- a/libretro.cpp
+++ b/libretro.cpp
@@ -80,6 +80,9 @@ bool cd_async = false;
 bool cd_warned_slow = false;
 int64 cd_slow_timeout = 8000; // microseconds
 
+// If true, PAL games will run at 60fps
+bool fast_pal = false;
+
 #ifdef HAVE_LIGHTREC
 enum DYNAREC psx_dynarec;
 bool psx_dynarec_invalidate;
@@ -3322,6 +3325,17 @@ static void check_variables(bool startup)
       widescreen_hack = false;
    }
 
+   var.key = BEETLE_OPT(pal_video_override);
+   if (environ_cb(RETRO_ENVIRONMENT_GET_VARIABLE, &var) && var.value)
+   {
+      bool want_fast_pal = strcmp(var.value, "fast_pal") == 0;
+
+      if (want_fast_pal != fast_pal) {
+         fast_pal = want_fast_pal;
+         has_new_timing = true;
+      }
+   }
+
    var.key = BEETLE_OPT(analog_calibration);
    if (environ_cb(RETRO_ENVIRONMENT_GET_VARIABLE, &var) && var.value)
    {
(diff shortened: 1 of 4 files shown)


libretro.cpp
```diff
@@ -4680,6 +4694,9 @@ void retro_deinit(void)
 
 unsigned retro_get_region(void)
 {
+   // simias: should I override this when fast_pal is set?
+   //
+   // I'm not entirely sure what's that used for.
    return content_is_pal ? RETRO_REGION_PAL : RETRO_REGION_NTSC;
 }
```
simias: Can somebody tell me if this is correct? I'm not sure what this is used for, as far as I can tell it's purely a frontend thing.

libretro_core_options.h
```diff
@@ -476,6 +476,16 @@ struct retro_core_option_definition option_defs_us[] = {
...
    },
+   {
+      BEETLE_OPT(pal_video_override),
+      "PAL (European) video timing override",
+      "Due to different standards PAL games often appear slowed down compared to the American or Japanese NTSC releases. This option can be used to override the timings in order to attempt to run these games with the NTSC framerate.",
+      {
+         { "no_override", "Standard PAL timings" },
+         { "fast_pal",  "PAL resolution, NTSC framerate" },
+      },
+      "no_override"
```
simias: I see that there's now a new file for internationalization, should it be updated as well or will it fail over to the English version for options that haven't been translated? Unfortunately I don't speak any of the currently implemented languages well enough to provide the translation...

libretro.cpp
ggdrt: Not sure what this is used for either, maybe it's a carryover from an older version of the libretro API?

libretro_core_options.h
ggdrt: The way it was explained to me in pull request https://github.com/libretro/beetle-psx-libretro/pull/608 is that internationalized core options work as overrides, so it is safe for us not to update the translations when adding new core options.

libretro.cpp
ggdrt: Asked around, and Autechre says it comes from back when libretro started off as a SNES API. He says he doesn't think it's used right now by RetroArch.

I'm not sure what old frontends are doing with the info being returned by this function.  My guess is something related to timing and aspect ratio, which would probably not work entirely properly with fast PAL hacking. I think it's probably safe to leave as-is with the comments provided until there's a demonstrable problem that occurs with this.

PR #638, opened 2020-04-09
Fix disk control interface when running single-disk PBP content

jdgleaver:
It turns out that the internal state variable `PBP_DiscCount` is set to zero when loading single-disk PBP content. This breaks the disk control interface, since the core behaves as though no disks are available - disk 'swapping' therefore behaves unexpectedly, and the frontend will always report a failure when trying to restore the last used disk.

This trivial PR fixes the issue.


AlexxandreFS:
I compiled this modification and it worked perfectly.
Thank you very much for the correction.

jdgleaver:
You're welcome - I'm glad to help!

PR #642, opened 2020-04-22
Improve DualShock calibration reference

simias:
I ran some tests on the real hardware and found that my original estimate was significantly lower than the real thing. A good test case is Parasite Eve because the radius it uses to consider that you're running instead of walking is very large and if you under-calibrate you'll end up walking when you go diagonally.

Beyond that, I wonder if it wouldn't make sense to turn this option on by default.

### DIFF
--- a/input.cpp
+++ b/input.cpp
@@ -975,9 +975,19 @@ void input_update(bool libretro_supports_bitmasks, retro_input_state_t input_sta
 
                   // This represents the maximal value the DualShock sticks can
                   // reach, where 1.0 would be the maximum value along the X or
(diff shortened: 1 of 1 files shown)


hizzlekizzle:
Yes, I've seen several people mention not being able to run on diagonals with Parasite Eve. Good catch.

Awakened0:
Ape Escape has that problem sometimes as well. Turning calibration on mostly fixes it, but I think it could also be improved by this.

simias:
Okay, I guess it's safe enough to merge as it is. We'll see if people complain about it...

PR #645, opened 2020-05-08
Dynarec updates

ZachCook:
Compile in support for lightrec by default. (HAVE_LIGHTREC=1 no longer required when compiling)

Update to latest lightrec and move PGXP support out of lightrec.c, now using unmodified lightrec.

Fix tracking/logging of mmap failures.


ZachCook:
LGTM's remaining new alerts (and large diff) are just from compiling with lightrec for the first time, and so finding the existing FIXME's in deps

PR #655, opened 2020-05-18
Cleanup warnings: snprintf truncation, strncpy->memcpy, memset

ZachCook:
Fix all build warnings other than those from libretro-common/ or deps/

Would like feedback on whether these changes make sense and don't break anything, as I just made the minimum changes to fix each warning and only build tested.


twinaphex:
These seem like good general purpose changes. I especially like getting rid of strncpy.

I will merge this tomorrow.

PR #659, opened 2020-05-20
Allow Solaris 11 build

kwyxz:
As title says, this allows the core to build (and run) on Solaris 11

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -88,8 +88,14 @@ endif
 ifneq (,$(findstring unix,$(platform)))
    TARGET := $(TARGET_NAME)_libretro.so
    fpic   := -fPIC
-   SHARED := -shared -Wl,--no-undefined -Wl,--version-script=link.T
-   ifneq ($(shell uname -p | grep -E '((i.|x)86|amd64)'),)
+   ifneq ($(findstring SunOS,$(shell uname -a)),)
+      GREP = ggrep
+      SHARED := -shared -z defs -z gnu-version-script-compat
+   else
+      GREP = grep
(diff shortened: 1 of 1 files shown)


niacat:
This makes the build fail on illumos because their version of the SunOS linker doesn't have gnu-version-script-compat.

I think it'd probably be better to leave that off.

kwyxz:
Is `-shared -z defs` still ok on IllumOS?

niacat:
Yeah, that should be fine

FWIW we've been shipping libretro builds for illumos for a while through pkgsrc, and our build system strips the GNU specific linker arguments out... but it doesn't know what to do with Oracle Solaris linker arguments

kwyxz:
@niacat after a quick audit of my repos I see the option in these
```
libretro-kronos/yabause/src/libretro/Makefile:		SHARED := -shared -z defs -z gnu-version-script-compat
libretro-mednafen_psx/Makefile:      SHARED := -shared -z defs -z gnu-version-script-compat
libretro-opera/Makefile:        SHARED := -shared -lpthread -lm -z defs -z gnu-version-script-compat
libretro-virtualjaguar/Makefile:		SHARED := -shared -z defs -z gnu-version-script-compat
libretro-vitaquake3/Makefile:      SHARED := -shared -z defs -z gnu-version-script-compat
libretro-xrick/Makefile.libretro:      SHARED := -shared -z defs -z gnu-version-script-compat
libretro-yabause/yabause/src/libretro/Makefile:		SHARED := -shared -z defs -z gnu-version-script-compat
```
I will be removing it everywhere.

PR #671, opened 2020-06-05
allow using system libs

alucryd:
This allows building this core against system libchdr and zlib, similar to my PR to flycast.

I have backported and upstreamed your chd_precache work to libchdr, with this I'm successfully using the beetle-psx cores with vanilla libchdr and zlib.

It would be awesome if you could upstream your ongoing CHD parent-clone work when it's done, upstream is very open to it https://github.com/rtissera/libchdr/issues/17.


twinaphex:
@alucryd I'm fine with merging this.

As for re-syncing libchdr from upstream, if the code can compile as C89 then yes we can do it. If there are some things that prevent a certain platform from compiling on our end, would upstream be receptive to merging PRs to solve this?

alucryd:
Thanks for merging! I'll let @rtissera answer your question :)

PR #683, opened 2020-06-29
Dynamically switch memory cards (fixes #658)

### LINKED ISSUE #658: [Feature Request] Dynamically switch memory cards
There are quite a few games, such as Xenogears, that allow you to save over multiple blocks in a memory card. I like to be able to maintain a backlog of saves so I can return to any point in the game or potentially collect anything I missed several hours back without restarting the game entirely. Of course, this means that I’d have to switch out lots of memory cards, especially for longer games.

As far as I can tell, memory cards are loaded and written when the core is opened and closed, so to switch memory cards I’d have to save in the right slot, close the core, switch the memory card in the left slot, and reopen the core. This can get a bit cumbersome (especially if the game only supports the left slot), so I was wondering if an option to dynamically switch out, or at least reload, the memory cards could be implemented.

rouhannb:
The issue (#658) has been collecting dust for over a month now, so I decided to make this. I won’t claim this code is good, efficient, or clean (I am an amateur), but as far as I can tell it gets the job done well.

(Also, what’s the rationale for not exposing `InputDevice_Memcard` in the header, and then having private methods for a class only used in one file? It looks unusual to me, so I am a bit curious.)


twinaphex:
OK, so couple of questions, is this optional? What exactly does it do?

rouhannb:
This allows the user to swap out memory cards while the core is loaded. Instead of always using 0.mcr for the left slot and 1.mcr for the right slot, they could choose to use any number (up to 64) for either slot. It’s as if they had a collection of memory cards, and physically took out the one in the console and replaced it with a different one. This would allow them to keep a backlog of saves that they could return to at any point without having to close the core and fiddle with the files manually.

I’m not sure what you mean by “optional”. If you’re asking if this would be unintrusive, that the user would not need to manually perform some action to return to how this worked before, then, yes, this is optional. By default, the left memory card will be saved to 0.mcr and the right card to 1.mcr, like normal, and it won’t do anything with the left card if it uses the Libretro/srm format.

negativeExponent:
> OK, so couple of questions, is this optional? What exactly does it do?

more file io access...

simias:
One important thing to do to emulate memorycard switching correctly is to make sure the `presence_new` flag in mednafen/psx/input/memcard.cpp is reset to `true` when switching to a new memory card. This flag is normally set by the hardware to let the game know that the Memory Card changed. I don't know if your patch does it but it's worth double checking.

Beyond that I considered such a feature in the past but I must admit that I'm not really sold on its utility. Juggling several Memory Cards was a pain on the real hardware and it's potentially even more of a pain in a libretro core IMO. If we could implement a nice interface that could list the contents of the Memory Cards before we load them that would be a different story, but it's not really easy to do in libretro as far as I know. If we could just list the titles of the saves somewhere, I think it would be a huge usability benefit.

rouhannb:
> One important thing to do to emulate memorycard switching correctly is to make sure the `presence_new` flag in mednafen/psx/input/memcard.cpp is reset to `true` when switching to a new memory card. This flag is normally set by the hardware to let the game know that the Memory Card changed. I don't know if your patch does it but it's worth double checking.

It didn’t do so before, but it does now.

> Beyond that I considered such a feature in the past but I must admit that I'm not really sold on its utility. Juggling several Memory Cards was a pain on the real hardware and it's potentially even more of a pain in a libretro core IMO.

How so? If this PR doesn’t go through I still plan on using this myself, so I’d appreciate it if you could be more specific so I know what I’m getting into. Is it just that it’d be hard to guess on which card a particular save is in?

simias:
Oh I think your implementation is as good as it can be, that wasn't really a criticism of your work. It's more of a limitation with what you can do in terms of console-specific UI with libretro.

The main problem IMO is that you can't really know what's on a given card before you load it. Then when you do, you have to wait for a while for the game to read the contents and display it to you (because memory cards are so damn slow). So if you have a bunch of memory cards and you're looking for a specific save but you don't know where it is, it's a slow and painful process to load every image one after the other.

Technically it could be possible for an emulator to display the contents of the card before loading it, so that the user can select the right one from the emulator's UI. Unfortunately I don't think there's a simple way to do something like that with libretro.

simias:
Regarding this last patch you made: the `presence_new` flag should be set even when loading a non-blank memory card: it's just a flag that tells the BIOS/game that a new memory card has been plugged in. Maybe I misunderstand what your code does but since I see a call to "Format" I suspect that your modification only does it for a blank card, not hot-switching valid images at runtime.

rouhannb:
Whoops, I should have realized that :/  
And yes, that is what my code does.

This should work now. (Hopefully.)

twinaphex:
@simias Are you in favor of merging this? Are there any downsides to it?

simias:
I'm not opposed to it (at least the concept behind it, I haven't taken the time to test the implementation).

As I mentioned previously the only downside I can think of is that if a clueless user starts tweaking options without taking the time to think about what they're doing they may end up thinking they lost their saves because the emulator would've switched to a different card. Given that beetle already had savestates and different ways to handle the memory cards themselves (libretro and mednafen) that creates a vast array of ways to mess up the config.

I can see how that can be useful though, so I guess it can't hurt merging it if only to get feedback?

twinaphex:
Alright, we'll merge it and if we hear any complaining, we'll have to contact @rouhannb again and figure out where to go from there.

Issue #648, opened 2020-05-08
[Beetle PSX HW][Vulkan] Filtering sprites in 3d games


EtSL33py:
## Description
Problem happens in games with 3D environment. Sprite alpha-channel not filtered or filtered incorrectly. With any filter, sprites have artifacts at the edges. I tried to pictured what is happening:
![vh2s](https://user-images.githubusercontent.com/19521365/81435210-2321be00-9170-11ea-945b-a12f0ee77b59.png)
### Actual behavior
![bof4](https://user-images.githubusercontent.com/19521365/81436222-c7f0cb00-9171-11ea-8075-99c6306371f5.png)
![fm3](https://user-images.githubusercontent.com/19521365/81436226-c9ba8e80-9171-11ea-9f38-3613a4c4b505.png)
![vh2](https://user-images.githubusercontent.com/19521365/81436229-cc1ce880-9171-11ea-91df-d9a22da9f27f.png)
![bof3](https://user-images.githubusercontent.com/19521365/81436235-cd4e1580-9171-11ea-9638-22f315b7c811.png)
### Expected behavior
This screenshots from opengl
![bof4gl](https://user-images.githubusercontent.com/19521365/81436354-0ab2a300-9172-11ea-8d58-47b2718d82ab.png)
![fm3gl](https://user-images.githubusercontent.com/19521365/81436364-0be3d000-9172-11ea-9200-2a7cb12e69c5.png)
![vh2gl](https://user-images.githubusercontent.com/19521365/81436368-0dad9380-9172-11ea-9181-b7df61d0a4d4.png)
![bof3gl](https://user-images.githubusercontent.com/19521365/81436371-0edec080-9172-11ea-8ef0-60a328d9b368.png)
### Version/Commit
Beetle PSX HW 0.9.44.1 9e0e95b
### Core options
Default options, only vulkan and any texture filter
### Environment information
- Frontend: Retroarch 1.8.5
- OS: Windows 10 1809 x64
- GPU: RX470 Adrenalin 20.4.1

hizzlekizzle:
This is a result of the way PS1 handles transparency. It's not like normal alpha, and this is actually the *improved* version.

see: 11e1db403381e4a19267b1af8281047280f43799

PR #695, opened 2020-07-27
Fixes for filtered texture rendering

amoser:
Fixes for issues highlighted in the attached "before" and "after" images.

before
after

### DIFF
--- a/parallel-psx/glsl/primitive.frag
+++ b/parallel-psx/glsl/primitive.frag
@@ -34,6 +34,10 @@ void main()
 		NNColor = sample_vram_atlas(clamp_coord(vUV));
 	}
 
+#if defined(OPAQUE) && defined(SEMI_TRANS) && defined(SEMI_TRANS_OPAQUE) 
+#error "Invalid defines."
+#endif
+
 	// Even for opaque draw calls, this pixel is transparent.
 	// Sample in NN space since we need to do an exact test against 0.0.
 	// Doing it in a filtered domain is a bit awkward.
@@ -41,12 +45,6 @@ void main()
 	// In this pass, only accept opaque pixels.
 	if (all(equal(NNColor, vec4(0.0))) || NNColor.a > 0.5)
 		discard;
-#elif defined(OPAQUE) || defined(SEMI_TRANS)
-	if (all(equal(NNColor, vec4(0.0))))
-		discard;
-
-#else
-#error "Invalid defines."
 #endif
 
 #ifdef SEMI_TRANS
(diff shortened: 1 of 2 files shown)


twinaphex:
Thank you, that looks like a big improvement! Well done!

hizzlekizzle:
This is great! Does it close this issue out, as well? https://github.com/libretro/beetle-psx-libretro/issues/648

backgamon:
For me it's not work :/
retroarch 2020-07-27 20-13-45-654
retroarch 2020-07-27 20-13-50-932

amoser:
> 
> 
> For me it's not work :/
> retroarch 2020-07-27 20-13-45-654
> retroarch 2020-07-27 20-13-50-932

Did you run the rebuild shaders script in the glsl folder? Just running make doesn't seem to capture those changes. 

Not that I'm ruling out the possibility that I made some kind of terrible mistake when making this pull request (like forgetting a file or something) -- let me triple check!

backgamon:
Rebuilding shaders make it work, good job!
At least now Psx Beetle are the best Psx emu so far :)
retroarch 2020-07-28 10-26-42-255

twinaphex:
@amoser Can you please commit these re-generated shaders? Unless I'm misunderstanding something, if it's the glslc-based shaders that parallel psx uses, users shouldn't really be expected to have to do that manually, the emulator should already have the latest up-to-date generated shaders in its core code.

amoser:
> 
> 
> @amoser Can you please commit these re-generated shaders? Unless I'm misunderstanding something, if it's the glslc-based shaders that parallel psx uses, users shouldn't really be expected to have to do that manually, the emulator should already have the latest up-to-date generated shaders in its core code.

I can do that for sure, if it helps! To be clear, my expectation certainly wasn't for users to need to do anything when just running a compiled binary of the emulator -- I had assumed (perhaps naively) that that the automated build system would also recompile any out-of-date shaders, and that manually running the rebuild_shaders.sh script would only be ever be necessary when building the entire core from source (which is what I assumed backgamon was doing). 

The fact that the pre-compiled shaders are included in the repo _along with_ their respective source code probably was a hint that they might need to be committed as well in order to stay up-to-date.

So if I'm understanding this correctly, I should make an additional commit which includes the _compiled_ versions of the modified shaders, which are located in "parallel-psx\glsl\prebuilt" . Sorry for any inconvenience, and please let me know if this still doesn't sound like what you're expecting.

twinaphex:
No need to apologize.

And yes, that sounds fine - just regenerate the shaders so the 'prebuilt' shaders are updated and committed in a separate PR.

amoser:
Done.

PR #713, opened 2020-09-02
Update JINC2 to prevent NaN, which result in black texture in some cases

xzn:
Discovered this when messing around with vertex shader to center pixel to screen coordinates. JINC2 filter would result in black textures in some cases.

This change resolves it.


rz5:
Could you please share some visual examples of what is fixed?

xzn:
Sure:

I'm modifying the vertex shader (for testing) to be like [Duckstation](https://github.com/stenzek/duckstation/blob/9254fc9e639af88cc972ac1c55d7b773b3afbecb/src/core/gpu_hw_shadergen.cpp#L534).

So the ground texture in this game looks normal:
image

Example of NaN when JINC2 is enabled:

image
image

This wouldn't happen before because vertex coordinate wasn't aligned with uv exactly like that. (I believe that's why anyhow)

If zero check is not handled per component wise then the dot product and other stuff that follows all result in NaN.

hizzlekizzle:
this looks good to me. Do we need to update the vulkan shaders accordingly, as well?

xzn:
Sorry totally forgot there's OpenGL as well. Added the change

PR #716, opened 2020-09-11
Use scaled_framebuffer/scaled_framebuffer_msaa for texture read if available

### LINKED ISSUE #704: [Vulkan] Internal resolution higher than 1x result in black outlines for certain 3D models of Vagrant Story
## Description

In Vagrant Story game uses sort of a trick to create illusion of DoF/Blur when object or character stands close to camera by, from what I can tell, duplicating it’s 3D model on black background into VRAM, slightly enlarging it and changing its transparency.
When using internal resolution of anything other than 1x you end up getting artifacts at the edges of the models due to the black background of the model not being removed properly.

### Expected behavior

Beetle PSX, internal resolution 4x, look at the right character "doubled" model, no artifacts
SW x4

### Actual behavior

Beetle PSX HW, internal resolution 4x, look at the right character "doubled" model, black outline artifacts
HW x4

1. Play Vagrant Story using Beetle PSX HW core
2. Watch cutscenes to witness black outline artifacts whenever fake DoF/Blur is shown

### Bisect results

When checking Display Full VRAM you can see the 3D model on the black background in the bottom right of the display whenever the object will be used for DoF/Blur effect

VRAM preview example 
Vagrant Story (USA)-200811-105509

Duckstation, different PSX core for the Retroarch using Vulkan as a renderer is not having any problems rendering the scene properly even at 8x the internal resolution.

Duckstation, internal resolution 8x
Vagrant Story (USA)-200811-094027

Changing other core options for Beetle PSX HW produces same results. Only internal resolution scaling is at fault.

### Version/Commit

Beetle PSX HW: (0.9.44.1 9f126e5)

### Core options

Renderer: Hardware
Internal GPU resolution: 4x
Dithering Pattern: OFF

Rest of the settings at default.

### Environment information

- Frontend: Retroarch 1.9.0 
- OS: Windows 7 SP1
- GPU: Nvidia GTX 970

xzn:
Fixes #704

80ffdfe 66cc554 af43eea Use spec constants to get down the number of shader combinations (may introduce stuttering, more obvious when validation is turned on)

8f07739 fixes the issue (except for msaa), though the quality is lower than duckstation

e90953d 2076679 higher quality fix, performance impact likely due to memory bandwidth

746aed7 fixes the issue for msaa

1eb54df f0e727c performance improvement (should be as fast or faster than before this pr for the most part)

Certain cutscenes in Vagrant Story have about 10% less framerate but at least MSAA is working on AMD now. And the low framerate is due to Texture Filtering being quite taxing (on memory bandwidth I suppose?) So turn that off will give good framerate again

The primary concern I have rn is stuttering. For me it only really shows up when validation is enabled but if it also happens generally then we should use precompiled spirv shaders without spec constant again. The only trouble is that this pr would need about 100+ more shader combinations:

- 3 transparency mode
- 6 filtering mode
- 4 scalings for msaa
- scaled and unscaled version of shaders (performance cost is too great if we only read from scaled_framebuffer, likely due to memory access, especially when Texture Filtering is enabled)

That alone is 144 shader combinations plus there's other stuff too.

I'm exploring using glslang to generate all the shader combinations rn in this [branch](https://github.com/xzn/beetle-psx-libretro/tree/glslang). Basically use a C++ program instead of shell script for more flexibility. The branch is a bit stale but it'll be worked on if needed.

Finally I didn't test this on many games. The main difficulty was getting the battle cursor in FF7 working due to `atlas.cpp` missing some scissor test and hazard flags (e90953d), also getting the transition screen in Silent Hill working. I very much like to know if this PR breaks any other things. If anyone's willing to test this it will be greatly appreciated in terms of many thanks from me.


twinaphex:
Alright, as long as you feel comfortable that you can resolve the stuttering bug, then I can merge this.

Just one small request - this standalone program that you would be writing that would replace the current script, is it possible it be written in C instead of C++ or do you insist it has to be that way?

twinaphex:
Awesome work on the rest BTW.

xzn:
Thank you for your kind words!

Not sure why this is merged already but, the standalone program would be linking against glslang for spirv generation anyway (basically doing what `glslc` currently is doing), so either way there would need to be a C++ compiler.

Is there any reason that you want it in C? I'm currently abusing initializer list like [this](https://github.com/xzn/beetle-psx-libretro/blob/6eb1b966366b78a57620430f4af8ae17bb9a559f/parallel-psx/glsl/rebuild_shaders.list.cpp)

Or would you prefer if I write a C program that calls `glslc` like the current shell script? I agree the C++ syntax I have is a bit unwieldy.

twinaphex:
Well C would be more portable down the line. If you have to write it in C++, is C++98 an option?

xzn:
glslang requires a C++11 compiler anyhow. For portability I could write it in C and make system call to `glslc` I guess..

On a related note would a Python script be better? Admittedly I don't really know how to approach this in C. Sorry :|

xzn:
BTW I just noticed that this pr breaks `scanout_to_buffer`, `scanout_vram_to_texture` and `scanout_vram_to_buffer` if msaa is enabled. Unbreaking it rn..

Issue #720, opened 2020-09-13
Valkyrie Profile - Location title headers are not displayed correctly on Vulkan/OpenGL


antonio-higino:
## Description

Beetle PSX HW - Location titles are not displayed at all on Vulkan and there's a slight visual glitch on OpenGL. Region is NTSC/USA.

### Expected behavior

- Vulkan (older build):

![VP_Working](https://user-images.githubusercontent.com/56856275/93029247-2f22fe00-f5f0-11ea-850c-cd0ae9686b68.png)

### Actual behavior

- Vulkan (latest build):

![VP_Bugged](https://user-images.githubusercontent.com/56856275/93029261-4530be80-f5f0-11ea-96b9-d7cb2bf6a10d.png)

- OpenGL (no change between older or latest build):

![VP_GL_Working](https://user-images.githubusercontent.com/56856275/93029938-73fd6380-f5f5-11ea-8c5a-bd52e5e0f89d.png)

### Steps to reproduce the bug

When using Vulkan/OpenGL, go to any place in the game (town, dungeon or other) that displays a location header, such as the screenshots above.

I could not reproduce the exact same issue on both backends. OpenGL displays the location's name correctly but showcases slightly different behavior from the older working build on Vulkan, notice the top half of the header's background rectangle is gray instead of white.

### Version/Commit

- Latest version: Beetle PSX HW 0.9.44.1 ab51d46

![Latest_Bugged](https://user-images.githubusercontent.com/56856275/93029387-f0417800-f5f0-11ea-81ce-3c4528180945.png)

- Old version: Beetle PSX HW 0.9.44.1 9f126e5

![Old_Working](https://user-images.githubusercontent.com/56856275/93029366-da33b780-f5f0-11ea-833e-ca572a692c6b.png)

### Core options

No core options or RetroArch options were changed between testing on the different versions.

- Core options used are:

beetle_psx_hw_adaptive_smoothing = "disabled"
beetle_psx_hw_analog_calibration = "disabled"
beetle_psx_hw_analog_toggle = "disabled"
beetle_psx_hw_aspect_ratio = "corrected"
beetle_psx_hw_cd_access_method = "sync"
beetle_psx_hw_cd_fastload = "2x(native)"
beetle_psx_hw_core_timing_fps = "force_progressive"
beetle_psx_hw_cpu_dynarec = "disabled"
beetle_psx_hw_cpu_freq_scale = "100%(native)"
[...]

PR #722, opened 2020-09-16
Scaled read (fix performance regression)

xzn:
Two changes:

ac16efd  avoid dividing by FB_SIZE which is slow, multiply by SCALE and use texelFetch instead.

b162646 [fixed a bug](https://github.com/libretro/beetle-psx-libretro/commit/b16264646dd494e8f97a7aed9f2ccf6b3a9df78c#diff-5dfb22247554410c870c7ddc6a7069aeR2013) where scaled_read variable was not updated in inner loop.

Should fix performance regression introduced by #716


twinaphex:
Let me know when this is ready to be merged.

xzn:
b162646 should fix #720 for the Vulkan case (my fault for the regression)

6124a06 fix some additional flickering, not sure what the performance penalties are; quick test doesn't show much impact.

xzn:
Something might as well discuss here:

Both [this](https://github.com/libretro/beetle-psx-libretro/blob/master/parallel-psx/renderer/renderer.cpp#L1747) and [this](https://github.com/libretro/beetle-psx-libretro/blob/master/parallel-psx/renderer/renderer.cpp#L1794) conditions are different than what's in [`select_pipeline`](https://github.com/libretro/beetle-psx-libretro/blob/master/parallel-psx/renderer/renderer.cpp#L1579)

Wouldn't there be cases where a primitive is queued for both `semi_transparent_opaque` and `semi_transparent`? Or is that how it's intended to work? From a quick glance the gl code and duckstation doesn't work this way.

Specifically if a primitive is both textured and semi-transparent, then it's put in both queue.

While on the subject what exactly is `semi_transparent_opaque`? How is it intended to be different from `semi_transparent`?

Pardon my confusion :o

xzn:
Also regarding this [comment](https://github.com/libretro/beetle-psx-libretro/pull/716#issuecomment-691122808),

The problem I have right now is that I'm generating about 10 to 20 megabytes of shader code. The code in `renderer.cpp` that will make use of all the different shader combinations will also gets quite complicated.

If I introduce compression to get down the binary size it'll complicate the code further.

Any good idea on how to proceed? (Introducing `glslang` as a runtime dependency could be an option but it also increases binary size, as well as load time)

xzn:
In my previous PR [this code](https://github.com/libretro/beetle-psx-libretro/blob/d057a30e74fcee86757084804ec199b2c8e2a757/parallel-psx/renderer/renderer.cpp#L1414) doesn't handle orientation at all. 

I made some changes in 8ecc2bc  to workaround this.

twinaphex:
> The problem I have right now is that I'm generating about 10 to 20 megabytes of shader code. The code in renderer.cpp that will make use of all the different shader combinations will also gets quite complicated.

> If I introduce compression to get down the binary size it'll complicate the code further.

> Any good idea on how to proceed? (Introducing glslang as a runtime dependency could be an option but it also increases binary size, as well as load time)

Hmm, let me propose this to someone who likely knows best how to proceed.

PR #725, opened 2020-09-18
"Clamp" pgxp coordinates if lacking w components

xzn:
I don't really know how to make PGXP less glitchy. This however adds a failsafe if a pgxp coordinate is lacking w component:

If the difference between pgxp x and y components and non-pgxp x and y components are too great, the non-pgxp values are used.

This in effect "hides" some more glaring glitches..

### DIFF
--- a/libretro.cpp
+++ b/libretro.cpp
@@ -126,7 +126,7 @@ enum dither_mode psx_gpu_dither_mode;
 
 //iCB: PGXP options
 unsigned int psx_pgxp_mode;
-unsigned int psx_pgxp_2d_tol;
+int psx_pgxp_2d_tol;
 unsigned int psx_pgxp_vertex_caching;
 unsigned int psx_pgxp_texture_correction;
 // \iCB
--- a/libretro.cpp
+++ b/libretro.cpp
@@ -126,6 +126,7 @@ enum dither_mode psx_gpu_dither_mode;
 
 //iCB: PGXP options
 unsigned int psx_pgxp_mode;
+unsigned int psx_pgxp_2d_tol;
 unsigned int psx_pgxp_vertex_caching;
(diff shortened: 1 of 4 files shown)


hizzlekizzle:
That's a good idea.

xzn:
The Spanish description is machine translated, sorry.. Maybe someone can help me check if the translation is indeed correct?

twinaphex:
Hi, not sure if this is related to this PR, but there is a major problem now -

if you turn PGXP off in a game like Tomb Raider, it renders nothing onscreen, just black. It only renders something if PGXP is turned on.

Is this something you can fix? I tested this with Vulkan.

twinaphex:
UPDATE: Graphics seem completely broken with PGXP off. The only way to get proper graphics output again is to turn PGXP on. Definitely worrying and should be looked at.

xzn:
I made some braindead mistake, fixing it asap :[

twinaphex:
No problem, and BTW it's fixed now! Thanks very much for the quick fix!

DonelBueno:
@xzn Do you know which games this fix?

xzn:
@DonelBueno This is only for when PGXP is set to CPU mode, certain UI elements no longer go off screen or stretched in a weird way.

The game intended to be improved was Silent Hill. With PGXP set to CPU mode the tree leaves have less jitter though the map screen would flicker when zooming.

Most games don't need CPU mode PGXP fortunately.

DonelBueno:
@xzn yeah, I know, I was just trying to discover more games where this option help.

I made a feature request for stenzek, from DuckStation, so he implements it there.

Nice to know about Silent Hill, I wasn't aware CPU Mode helped it. From my testing, it also helps Wipeout XL and Looney Tunes Racing when using CPU Mode.

Issue #490, opened 2019-02-26
[Beetle PSX HW] Silent hill cutscene with PGXP operation mode


Brh24:
This is a issue that has already been reported, when reach a specific part in the game after a cutscene, during the dialogue the game freeze altogether example in the picture below.

![silent hill usa -190226-192141](https://user-images.githubusercontent.com/48028906/53436677-07634980-39fc-11e9-8778-a2dcc22fca23.png)

Now apparently this happen only if you have PGXP operation mode set with memory only.

I'm using 1.76 Beetle PSX HW with vulkan render, my specs are the following
windows 10 64 bit, amd ryzen 5 1500 quadcore,  8gb ram,  nvdia gtx 1050

twinaphex:
@iCatButler This might be an interesting edge case issue for you.

rz5:
@Brh24 - Do you have a savestate of the game right before that bug happens? It would be helpful if you could share it here.

Brh24:
[Silent Hill (USA).zip](https://github.com/libretro/beetle-psx-libretro/files/2910254/Silent.Hill.USA.zip)
this should be it @rz5

iCatButler:
This is related to PGXP intercepting the GTE's NCLIP operation and allowing triangles that would normally be degenerate to pass through and be rendered instead of culled. If I prevent this from happening then the cutscene plays as intended.

I think the actual root cause of the error is in the game itself not being able to handle the extra geometry rendering as similar corruption occurs if I enable the widescreen hack, with or without PGXP. It is also fixed if I set NCLIP to simply reject every other triangle, regardless of whether it would be culled or not.

In this case the game may be allocating a fixed size buffer to store all the draw calls for each frame (with double buffering) but not checking to see if these buffers overflow when draw calls are written into them. If one buffer overflows into the other it could corrupt the DMA chain currently being read by the GPU. Normally there would be no single scene with enough visible geometry to cause this, but with the widescreen or PGXP enabled it happens in this specific scene.

I'd be interested to know if this happens at all in PCSX-R or ePSXe, as their DMA chain behaviour is different. The Pete's GPU plugins always try to process all nodes in the buffer without passing back execution to the CPU emulation, whereas Mednafen seems to  alternate cycles on each to simulate them working in parallel.

PukeGX:
Just wanted to add that the same problem occured with the software Beetle PSX with all default options, just widescreen hack on

Papermanzero:
[...]

Issue #586, opened 2020-01-17
[Vulkan] "Tobal No.1 (USA)" palette scanline is spammed to fill display


rz5:
## Description
The Vulkan renderer has a bug where after the first "NOW LOADING" screen, the second scanline, which seems to contain texture color palettes, is repeated to fill the display. The display will stay like that on menus and in-game, with only the second scanline getting updated. FMVs do display correctly.

### Expected behavior

Using the software and GL renderers, you get this:
![Tobal No 1 (USA)-200117-190408](https://user-images.githubusercontent.com/8117159/72639039-ae2a3e80-395c-11ea-8ac1-8ad4082677f5.png)

### Actual behavior

Using the Vulkan renderer, we get this:
![Tobal No 1 (USA)-200117-190251](https://user-images.githubusercontent.com/8117159/72639290-27c22c80-395d-11ea-9a5c-b5c4052c3d9a.png)

### Steps to reproduce the bug

1) Get past the intro FMV.
2) The display is now bugged as per the description.

### Bisect Results

I went as far back as 5a3093be0fb24dbb1d95303ae615ad4647cb54f6, which is before ggdrt's vulkan changes, maister's bug fixes, MSAA features, Granite updates and the bug was still happening. Perhaps the bug was always there. @twinaphex - if you remember some moment where it worked, please says so. Maybe it's a vulkan driver regression?

### Core options
All core options set to default.

### Environment information

- OS: Microsoft Windows [Version 10.0.17763.973]
- Compiler: gcc version 9.2.0 (Rev2, Built by MSYS2 project)
- GPU: Nvidia GTX 970, driver version 441.87

backgamon:
With a smallest "modified" version i don't have that issue:
![Screenshot - 17_01_2020 , 20_43_26](https://user-images.githubusercontent.com/39404647/72641335-35ca7a00-396a-11ea-9a7d-d8fdda8bfb8f.jpg)

rz5:
@backgamon You mind sharing what modifications you have on that build? Are you sure you're using the Vulkan renderer?

backgamon:
As i said on another issue, i dislike the new scaling option.
I'm not programmer bit i tryed many time to have my fullscreen scaling.
Then (with some help), i very basically learned how to compile(with Msys2), and i just removed ALL the work about new scaling.
Used a old version and added all new commit except from overscan.
Now i can use my fullscreen option without issue...
Actually i retest many games on my website.
[...]

PR #732, opened 2020-09-22
Add PGXP NCLIP on off switch

xzn:
Add option to turn off PGXP's NCLIP implementation. This is a workaround for #490.

Not sure if it's a good idea to add yet another option though this one is in duckstation.

(I spent some time following the suggestion iCB mentioned [here](https://github.com/libretro/beetle-psx-libretro/issues/490#issuecomment-469062488): first a snapshot of the memory is taken when the dma starts then the dma is fed from the snapshot instead of updated ram, could be my implementation being wrong but the game is still locking up in the same place.. So far I've no idea how to workaround this for the widescreen case)

### DIFF
--- a/libretro.cpp
+++ b/libretro.cpp
@@ -129,6 +129,7 @@ unsigned int psx_pgxp_mode;
 int psx_pgxp_2d_tol;
 unsigned int psx_pgxp_vertex_caching;
 unsigned int psx_pgxp_texture_correction;
+unsigned int psx_pgxp_nclip;
 // \iCB
 
 #define NEGCON_RANGE 0x7FFF
@@ -3390,6 +3391,17 @@ static void check_variables(bool startup)
       psx_pgxp_texture_correction = PGXP_MODE_NONE;
    // \iCB
 
+   var.key = BEETLE_OPT(pgxp_nclip);
+   if (environ_cb(RETRO_ENVIRONMENT_GET_VARIABLE, &var) && var.value)
+   {
+      if (strcmp(var.value, "disabled") == 0)
+         psx_pgxp_nclip = PGXP_MODE_NONE;
+      else if (strcmp(var.value, "enabled") == 0)
+         psx_pgxp_nclip = PGXP_NCLIP_IMPL;
+   }
+   else
+      psx_pgxp_nclip = PGXP_MODE_NONE;
+
    var.key = BEETLE_OPT(line_render);
    if (environ_cb(RETRO_ENVIRONMENT_GET_VARIABLE, &var) && var.value)
    {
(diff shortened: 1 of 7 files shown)


i30817:
I suspect this would be cleaner with a game patch, which beetle does not have the infrastructure for (like dolphin does i guess).

xzn:
Does beetle have a debugger like dolphin/pcsx2? I've been staring at `cpu.cpp` to figure out why PGXP is kinda wonky in Valkyrie Profile and haven't seen sign of a debugger. I'm not sure if I have the skill to write one myself..

(The rsx component has something called rsx_dump which I haven't figured out how to use either)

twinaphex:
Hi @xzn , it seems the Tobal No. 1 issue with the Vulkan renderer is back again. Can you take a look please to see what caused the regression?

xzn:
What exactly is the issue? Are the screens filled with vertical lines again? (Sorry I just tried the game and couldn't see any problems..)

twinaphex:
Vertical lines. I will try compiling from source in about 6 hours instead of the build from the buildbot to see if theres an issue on latest.

i30817:
> Does beetle have a debugger like dolphin/pcsx2? I've been staring at `cpu.cpp` to figure out why PGXP is kinda wonky in Valkyrie Profile and haven't seen sign of a debugger. I'm not sure if I have the skill to write one myself..
> 
> (The rsx component has something called rsx_dump which I haven't figured out how to use either)

Mednafen does https://mednafen.github.io/documentation/debugger.html , so if beetle doesn't it was either ripped out or disabled in the makefile...

xzn:
Does beetle compiles with `WANT_DEBUGGER` ? `mednafen/psx/debug.cpp` seems to be needing some functions that were in `mednafen/debug.cpp` which is not included in beetle. Maybe there was debugger functionality at one point, just need to figure out a way to get it working again.

With gdb I'm not sure how it'd work? For instance how do we set breakpoint in the game that is being emulated? Conditional breakpoint on the PC variable location could work i guess but it's an extra layer of indirection.

i30817:
Ah i guess you're right.

twinaphex:
@xzn It's definitely broken again on the latest version. The only way I can 'fix it' is by going to Core Options -> Internal Resolution and changing it so it triggers a video restart. However, obviously we shouldn't have to do this and it would just work as normal as it seems it did a few commits before. 

Would appreciate if you could investigate this and find out what broke it again like this.

xzn:
I really cannot reproduce it for some reason. I assume you are using the default settings? What platform are you running beetle on? Is the game broken like in [these](https://user-images.githubusercontent.com/49579368/84579250-929b5680-ad89-11ea-84f4-25148a5d49e5.png) [screenshots](https://user-images.githubusercontent.com/8117159/72639290-27c22c80-395d-11ea-9a5c-b5c4052c3d9a.png)?

I'm only testing it on Windows with a non-RTX turing card. I'm not sure if that makes any difference.. :[

rz5:
For reference, the "Tobal 1 issue" is tracked in #586. As far as I know, it was never fixed, but I have not been paying attention to @xzn's changes.

xzn:
I tried fixing it [here](https://github.com/libretro/beetle-psx-libretro/pull/723). It worked for some settings but it's no longer working now according to Autechre.

Update: I tested a bit more, adaptive smoothing is still not fixed so I'll need to resolve that.

@Autechre what other settings are you using that made this not working? I'd like to actually resolve them all this time.

Issue #746, opened 2020-10-14
Graphic issue after latest changes


Kupo91:
**Edit:** I tested the old core versions from the buildbot. The last working core is the one from 27th september. The issue occured for the first time with the core from 28th september.

________________________________________________________

I just updated my core a few days ago. After updating the character sprites in certain positions have these extra pixel strips which don't belong there. I'm not good at explaining so here are some screenshots:

![Grandia  D SLES-02399 -201014-104623](https://user-images.githubusercontent.com/41426906/95967523-a4126f00-0e0c-11eb-8277-4337b82f778d.png)
![Grandia  D SLES-02399 -201014-104809](https://user-images.githubusercontent.com/41426906/95967535-a70d5f80-0e0c-11eb-81e1-13e0fc573501.png)
![Grandia  D SLES-02399 -201014-105123](https://user-images.githubusercontent.com/41426906/95967541-a8d72300-0e0c-11eb-90eb-27b3748e2c8d.png)
![Grandia  D SLES-02399 -201014-105159](https://user-images.githubusercontent.com/41426906/95967545-aaa0e680-0e0c-11eb-871d-f5dfb32de802.png)
![Grandia  D SLES-02399 -201014-105213](https://user-images.githubusercontent.com/41426906/95967551-ac6aaa00-0e0c-11eb-86fa-5901862e56c0.png)

The Game is Grandia and I think you can see these pixel strips only in battles and if internal resolution is atleast x2.

I restored an old core from 2nd of july and this core does not have this issue! So it seems like some of the latest changes causes this problem.

- Frontend: Retroarch Oct 12 2019
- OS: Linux (v5.3)
- GPU: Intel Iris Plus Graphics 655, video driver is vulkan

bslenul:
I can reproduce on Windows 10 also, bisect done, first bad commit is 6f8751a.

barbudreadmon:
I see a new core option "Texture UV Offset" with the description "Sample textures for 3D polygons at an offset for higher than 1x internal resolution. Reduce texture seams but may cause unintended graphical glitches." in that commit, it's enabled by default so did you try disabling it ? (imho that's probably something that should be disabled by default if it's causing gfx issues)

bslenul:
I just tried turning it OFF (then close content just to be sure) and the same thing still happens for me.

twinaphex:
Hi @xzn , can you check out this regression? Thanks in advance.

xzn:
[...]

PR #748, opened 2020-10-15
Remove code from previous changes causing problems in Grandia

xzn:
For #746

I was hoping this code wouldn't cause problems like this but apparently it does. (Valkyrie Profile is just kind of hard to get right, maybe I'll try a different approach in the future but causing glitches in other games as pointed out is just unacceptable)

Removing the offending code is better than reverting the various related commits I suppose.

Apologies in advance and thanks to @bslenul for the bisect.

### DIFF
--- a/mednafen/psx/gpu.h
+++ b/mednafen/psx/gpu.h
@@ -150,7 +150,6 @@ struct PS_GPU
    // primitive UV limits (used to clamp texture sampling)
    uint16_t min_u, min_v, max_u, max_v;
    bool may_be_2d;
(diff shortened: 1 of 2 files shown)


twinaphex:
Ah thanks. I will merge this.

Hopefully this won't demotivate you from looking at implementing an alternative solution in the long run though.

bslenul:
Hey, no problem for the bisect! 👍 

For Valkyrie Profile, can that specific block be active only when playing that game? Or is this considered too "hacky"?

xzn:
It's literally a hack since it glitches out other games :(

PR #775, opened 2021-01-29
Added option to use shared memory card with libretro save  method (right card only)

housello:


housello:
Found a bug initialising memcard with some settings.
I cancel the pull request, will make another after fixing.

PR #782, opened 2021-03-22
Update libchdr

jdgleaver:
This PR updates libchdr, removing the dependency on libflac.

The PR also fixes two minor bugs, encountered while testing these changes with ASAN:

- A minor disk-image-related memory leak has been removed (a read thread condition wasn't being deleted)
- The dualshock `analog_mode` variable is now correctly initialised (previously we were reading an uninitialised value at startup)


twinaphex:
Thanks a lot!

PR #791, opened 2021-05-07
Dynarec updates

ZachCook:
Update to most recent versions of lightrec and lightning, and move their generated .h files to a separate directory to keep the deps directories as clean subrepos that can be updated easily without any manual merging.


ZachCook:
Would like some testing before merge on x86_32 and x86_64 due to updates in lightning code generation that should be checked for any remaining regressions, they previously caused falling through ground issues in Spyro and Crash for example, before being fixed in lightrec.

twinaphex:
Hi there, I can test x86_64 at least on Windows. Not sure if I have a 32-bit x86 machine setup somewhere but I'll have a look around.

pcercuei:
I think you can bump Lightrec to the 0.4 version.

ZachCook:
> I think you can bump Lightrec to the 0.4 version.
@pcercuei commit 29e0c90 in this PR does exactly that, I see that you have done some large changes since 0.4 though, do you think those are stable and important enough to add to this PR?

pcercuei:
Alright, I didn't see it since there are two commits to update Lightrec. Why not merge them?

The commits since 0.4 help to greatly reduce the blocks compilation time, so it becomes much less choppy when the dynarec has to compile lots of blocks. Definitely good to have. I didn't see any regressions with the games I tested, but there might be some.

ZachCook:
I updated to today's lightrec, squashed and moved some commits, finally got rid of those windows specific workarounds and it appears to be working fine, I'll be testing as much as I can, if nothing shows up in the next day or two we can probably merge this

ZachCook:
I was clearly too tired when testing earlier, removing the windows specific workarounds does break stuff, will undo

ZachCook:
I think we can merge without testing on older android versions, the additional fallback path for Android 10+ does not occur there, and even if it did it would fallback to non-mmap memory and still work, there are no changes specific to ARM version either, we already included the patch that affected aarch64 in gnu-lightning too.

PR #799, opened 2021-06-07
add loongarch support

loongson-zjl:
Signed-off-by: zhangjialing <zhangjialing@loongson.cn>


xen0n (changes requested):
This change is not needed (consider x86 for example), most Linux platforms are going to fallback to the last branch with `clock_gettime`. This syscall is certainly usable on LoongArch, and most probably with nanosecond precision, so even better than the MIPS branch's `gettimeofday` which only has microsecond resolution.

loongson-zjl:
> This change is not needed (consider x86 for example), most Linux platforms are going to fallback to the last branch with `clock_gettime`. This syscall is certainly usable on LoongArch, and most probably with nanosecond precision, so even better than the MIPS branch's `gettimeofday` which only has microsecond resolution.

Ha, I will close this pull requests

PR #800, opened 2021-06-10
Disable pre-cache for physical media

### LINKED ISSUE #795: [Bug/Workaround Request] Disable precaching when loading from physical media
Loading games via physical media stalls eternally when precache is enabled. Auto-disabling precache whenever the loaded file is named `drive[n].cue` would sort this nicely. In my case I have two drives, so I've had to make configs disabling it for `drive2` and `drive3`.

ZachCook:
This takes way too long and is unresponsive, so don't actually
pre-cache any physical media when pre-cache is enabled

Fixes #795

### DIFF
--- a/libretro.cpp
+++ b/libretro.cpp
@@ -3888,7 +3888,14 @@ static bool MDFNI_LoadCD(const char *devicename)
 
          image_label[0] = '\0';
 
-         CDIF *image  = CDIF_Open(&success, devicename, false, cdimagecache);
+         bool cache = cdimagecache;
+         /* don't precache if physical cdrom, will take way too long and be unresponive */
+         if (cdimagecache && devicename && !strncasecmp(devicename, "cdrom:", 6)) {
+            cache = false;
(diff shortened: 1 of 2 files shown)


twinaphex:
Umm, is this going to mean that pre-caching of games doesn't work in general? I find this to be a very handy feature when you want a game loaded into RAM when the storage it's on is too slow for it to normally playback and you don't want any disk I/O afterwards. On systems with a large enough pool of RAM this is not an issue, others should simply not use it.

ZachCook:
> Umm, is this going to mean that pre-caching of games doesn't work in general? I find this to be a very handy feature when you want a game loaded into RAM when the storage it's on is too slow for it to normally playback and you don't want any disk I/O afterwards. On systems with a large enough pool of RAM this is not an issue, others should simply not use it.

@twinaphex this only disables pre-cache when running a disc from a cd drive since they take about an hour to pre-cache with the current cd reading code, you can't even access the menu, it will seem like it just froze completely. Pre-cache will still work for any other source such as an already backed up disc. For users that always have pre-cache enabled this should prevent them thinking that cds don't work at all.

twinaphex:
Alright, I understand now.

PR #801, opened 2021-07-03
Adding Spanish translation to the core.

IlDucci:
Also updating the word "(Restart)" so it matches the current unified term on US texts.


DisasterMo:
@IlDucci can be closed now (#831)

PR #812, opened 2021-08-24
Dynarec updates, and revert my bad gpu optimization

ZachCook:
Actually fixes the Ape Escape crash and re-enable the optimization that caused it, which should restore any lost performance provided by that optimization

Some small lightrec memory savings

print a bit more info from lightrec on startup/shutdown: default LIGHTREC_LOG_LEVEL increased to 3 (INFO), notably the return of lightrec memory usage on shutdown.

Reverts d8479c7 due to a reduction in SW renderer performance as noted in #794, better to fix performance now and worry about trying to optimize for compile time and size later.


ZachCook:
I'm going to see if I can integrate lightrec's log messages into libretro ones

Issue #719, opened 2020-09-13
Auto crop for PAL games.


backgamon:
Since https://github.com/libretro/beetle-psx-libretro/commit/706a4e6f9c81f32564112d69e8731a2899399ea4 black bars show on many PAL games.
A option are added for cropping, but many need a differents values for each.
It almost never be perfect...

### Expected behavior

No black bars same as https://github.com/libretro/beetle-psx-libretro/commit/79935274c2f990a962a8c6a111fb1f49f7bf1d46.

### Steps to reproduce the bug

Just launch any PAL games on latest build.

### Bisect results

I compiled a old version with some of the new change (XBR and textures correcting), but i'm unable to add all changes like Lightrec and RSX...
There are any chances to add a option/switch to disable (or enable) a system like the older?
Same as Duckstation:
![crop](https://user-images.githubusercontent.com/39404647/93019833-d0eb2080-f5d9-11ea-895e-9181ed769b16.jpg)

### Environment information

Retroarch 1.9.0
Windows 10 Pro x64 [19041.508]
Gigabyte 2070 RTX 8G OC [452.06]

IngwiePhoenix:
Definitively having the same issue. I switched from DuckStation to this core - because it actually supports SBI files. DuckStation does not. Many of my PAL rips actually need that (Crash Bash being one of these).

I enabled the frame display. It fits pretty snug into that border, ironically.
![image](https://user-images.githubusercontent.com/2893716/93543761-bd82e100-f95c-11ea-9783-9041b0d45ba5.png)

An option to get rid of those would be nice. That with 16:9 and upscaled internal renderer is pretty amazing.

But those bars...kinda annoy me. My TV is an old 55" from 2011 and already has thick bezels - I don't need any more than I already have. :)

backgamon:
For who want a old version[18 october2019] with all correction about XBR and Corrupted texture:

https://mega.nz/file/92hhxBCD#sBf0N6M9-H7K5xML0AZkvjBjCouvtNGhn4050YzAocc

Lightrec and all new change (except correction) are unavailable...i can't add...

ds22x:
I believe this has been solved now.

PR #819, opened 2021-09-29
Make "Crop Overscan" dynamically crop (and center) the screen

ds22x:
In particularly inspired by https://github.com/libretro/beetle-psx-libretro/issues/719, this makes it so the Crop Overscan option dynamically crop and center the screen based on how PCSX-ReArmed and DuckStation deal with it (and especially PAL games, which have varying amount of crop, sometimes even in the same game, which makes manually adjusting the overscan and image offset values painful).
This also makes it that, if Crop Overscan is disabled, it will revert back to the unmodified image, including the varying PAL overscans.

### DIFF
--- a/libretro_core_options.h
+++ b/libretro_core_options.h
@@ -1091,9 +1091,9 @@ struct retro_core_option_v2_definition option_defs_us[] = {
    },
    {
       BEETLE_OPT(crop_overscan),
-      "Crop Horizontal Overscan",
+      "Crop Overscan",
       NULL,
-      "By default, the renderers add horizontal padding (pillarboxes on either side of the image) to emulate the same black bars generated in analog video output by real PSX hardware. Enabling this option removes horizontal padding.",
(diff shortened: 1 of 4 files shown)


hacls1:
Hi guys, there is a problem with the new Crop Overscan, before, which was more accurate, you could just chop ~8px top and bottom, that's by default the configuration of DuckStation, for example, and there you have it.

Now you pretty much crop all overscan independent of how much it is, FFIX is 8px, MGS is another, Chrono Cross another, etc.. Which is great, the problem is there's a ton of games that still display trash pixels, like Dino Crisis 2, Legend of Dragoon, etc and automatically cuting off the horizontal blackness (vertical overscan) leaves you having to config additional cropping game by game, which is a nightmare.

I suggest leaving the original option there and put some "PAL optimize" option or something like that for someone who needs it, like the "Crop Mode>All Borders" from DuckStation.

Exemplo1
Exemplo2

PR #833, opened 2021-12-05
Core options text adjustments

DisasterMo:
@IlDucci please review this, okay?

### DIFF
--- a/libretro_core_options.h
+++ b/libretro_core_options.h
@@ -65,12 +65,12 @@ struct retro_core_option_v2_category option_cats_us[] = {
    {
       "input",
       "Input",
-      "Change lightgun, mouse and neGcon settings."
+      "Change light gun, mouse and neGcon settings."
    },
    {
       "memcards",
       "Memory Card",
-      "This lets you modify settings related to the virtual Memory Card(s) used by the system."
+      "Change settings related to the virtual Memory Card(s) used by the system."
    },
    {
       "pgxp",
@@ -90,7 +90,7 @@ struct retro_core_option_v2_definition option_defs_us[] = {
       BEETLE_OPT(internal_resolution),
       "Internal GPU Resolution",
       NULL,
-      "Set internal resolution multiplier. Resolutions higher than '1x (Native)' improve fidelity of 3D models at the expense of increased performance requirements. 2D elements are generally unaffected by this setting.",
+      "Choose internal resolution multiplier. Resolutions higher than '1x (Native)' improve fidelity of 3D models at the expense of increased performance requirements. 2D elements are generally unaffected by this setting.",
       NULL,
       "video",
       {
@@ -108,7 +108,7 @@ struct retro_core_option_v2_definition option_defs_us[] = {
       BEETLE_OPT(depth),
       "Internal Color Depth",
       NULL,
-      "Set internal color depth. Higher color depth can reduce color banding effects without the use of dithering. 16 bpp emulates original hardware but may have visible banding if dithering is not enabled. 'Dithering Pattern' is recommended to be disabled when this option is set to 32 bpp.",
+      "Choose internal color depth. Higher color depth can reduce color banding effects without the use of dithering. 16 bpp emulates original hardware but may have visible banding if dithering is not enabled. 'Dithering Pattern' is recommended to be disabled when this option is set to 32 bpp.",
       NULL,
       "video",
       {
@@ -124,7 +124,7 @@ struct retro_core_option_v2_definition option_defs_us[] = {
       BEETLE_OPT(dither_mode),
       "Dithering Pattern",
       NULL,
-      "Set dithering pattern configuration. '1x (Native)' emulates native low resolution dithering used by original hardware to smooth out color banding artifacts visible at native color depth. 'Internal Resolution' scales dithering granularity to the configured internal resolution for cleaner results. Recommended to be disabled when running at 32 bpp color depth. Note: On Vulkan, enabling this will force downsampling to native color depth, while disabling will automatically enable output at higher color depth.",
+      "Choose dithering pattern configuration. '1x (Native)' emulates native low resolution dithering used by original hardware to smooth out color banding artifacts visible at native color depth. 'Internal Resolution' scales dithering granularity to the configured internal resolution for cleaner results. Recommended to be disabled when running at 32 bpp color depth. Note: On Vulkan, enabling this will force downsampling to native color depth, while disabling will automatically enable output at higher color depth.",
       NULL,
       "video",
       {
@@ -140,7 +140,7 @@ struct retro_core_option_v2_definition option_defs_us[] = {
       BEETLE_OPT(scaled_uv_offset),
       "Texture UV Offset",
       NULL,
-      "Sample textures for 3D polygons at an offset for higher than 1x internal resolution. Reduce texture seams but may cause unintended graphical glitches.",
+      "Sample textures for 3D polygons at an offset for higher than 1x internal resolution. Reduces texture seams but may cause unintended graphical glitches.",
       NULL,
       "video",
       {
@@ -156,7 +156,7 @@ struct retro_core_option_v2_definition option_defs_us[] = {
       BEETLE_OPT(filter),
       "Texture Filtering",
       NULL,
-      "Select texture filtering method. 'Nearest' emulates original hardware. 'Bilinear' and '3-Point' are smoothing filters, which reduce pixelation via blurring. 'SABR', 'xBR', and 'JINC2' are upscaling filters that may improve texture fidelity/sharpness at the expense of increased performance requirements. Only supported by the hardware renderers.",
+      "Choose texture filtering method. 'Nearest' emulates original hardware. 'Bilinear' and '3-Point' are smoothing filters, which reduce pixelation via blurring. 'SABR', 'xBR', and 'JINC2' are upscaling filters that may improve texture fidelity/sharpness at the expense of increased performance requirements. Only supported by the hardware renderers.",
       NULL,
       "video",
       {
@@ -175,16 +175,16 @@ struct retro_core_option_v2_definition option_defs_us[] = {
       BEETLE_OPT(filter_exclude_sprite),
       "Exclude Sprites from Filtering",
       NULL,
-      "Do not apply texture filtering to sprites. Prevent seams in various games with 2D sprite-rendered backgrounds. Use together with Adaptive Smoothing or another post-processing filter for best effect.",
+      "Do not apply texture filtering to sprites. Prevents seams in various games with 2D sprite-rendered backgrounds. Use together with Adaptive Smoothing or another post-processing filter for best effect.",
       NULL,
       "video",
       {
-         { "disable", NULL },
+         { "disabled", NULL },
          { "opaque", "Opaque Only" },
          { "all", "Opaque and Semi-Transparent" },
          { NULL, NULL },
       },
-      "disable"
+      "disabled"
    },
    {
       BEETLE_OPT(filter_exclude_2d_polygon),
@@ -194,12 +194,12 @@ struct retro_core_option_v2_definition option_defs_us[] = {
       NULL,
       "video",
       {
-         { "disable", NULL },
+         { "disabled", NULL },
          { "opaque", "Opaque Only" },
          { "all", "Opaque and Semi-Transparent" },
          { NULL, NULL },
       },
-      "disable"
+      "disabled"
    },
 #endif
 #endif
@@ -208,7 +208,7 @@ struct retro_core_option_v2_definition option_defs_us[] = {
       BEETLE_OPT(adaptive_smoothing),
       "Adaptive Smoothing",
       NULL,
-      "When enabled, smooths 2D artwork and UI elements without blurring 3D rendered objects. Only supported by the Vulkan renderer.",
+      "Smooth out 2D artwork and UI elements without blurring 3D rendered objects. Only supported by the Vulkan renderer.",
       NULL,
       "video",
       {
@@ -222,7 +222,7 @@ struct retro_core_option_v2_definition option_defs_us[] = {
       BEETLE_OPT(super_sampling),
       "Supersampling (Downsample to Native Resolution)",
       NULL,
-      "When enabled, downsamples rendered content from upscaled internal resolution down to native resolution. Combining this with higher internal resolution multipliers allows for games to be displayed with anti-aliased 3D objects at native low resolution. Produces best results when applied to titles that mix 2D and 3D elements (e.g. 3D characters on pre-rendered backgrounds), and works well in conjunction with CRT shaders. Only supported by the Vulkan renderer. Note: 'Dithering Pattern' is recommended to be disabled when enabling this option.",
+      "Downsample rendered content from upscaled internal resolution down to native resolution. Combining this with higher internal resolution multipliers allows for games to be displayed with anti-aliased 3D objects at native low resolution. Produces best results when applied to titles that mix 2D and 3D elements (e.g. 3D characters on pre-rendered backgrounds), and works well in conjunction with CRT shaders. Only supported by the Vulkan renderer. Note: 'Dithering Pattern' is recommended to be disabled when enabling this option.",
       NULL,
       "video",
       {
@@ -236,11 +236,11 @@ struct retro_core_option_v2_definition option_defs_us[] = {
       BEETLE_OPT(msaa),
       "Multi-Sampled Anti Aliasing",
       NULL,
-      "Set MSAA level for rendered content. Improves the appearance of 3D objects. Only supported by the Vulkan renderer.",
+      "Choose MSAA level for rendered content. Improves the appearance of 3D objects. Only supported by the Vulkan renderer.",
       NULL,
       "video",
       {
-         { "1x",  "1x (Default)" },
+         { "1x",  NULL },
          { "2x",  NULL },
          { "4x",  NULL },
          { "8x",  NULL },
@@ -253,7 +253,7 @@ struct retro_core_option_v2_definition option_defs_us[] = {
       BEETLE_OPT(mdec_yuv),
       "MDEC YUV Chroma Filter",
       NULL,
-      "Improves the quality of FMV playback by reducing 'macroblocking' artifacts (squares/jagged edges). Only supported by the Vulkan renderer.",
+      "Improve the quality of FMV playback by reducing 'macroblocking' artifacts (squares/jagged edges). Only supported by the Vulkan renderer.",
       NULL,
       "video",
       {
@@ -282,7 +282,7 @@ struct retro_core_option_v2_definition option_defs_us[] = {
       BEETLE_OPT(dump_textures),
       "Dump Textures",
       NULL,
-      "Dumps used textures to <cd>-texture-dump/",
+      "Dump used textures to <cd>-texture-dump/",
       NULL,
       "video",
       {
@@ -297,7 +297,7 @@ struct retro_core_option_v2_definition option_defs_us[] = {
       BEETLE_OPT(replace_textures),
       "Replace Textures",
       NULL,
-      "Replaces textures using HD versions from <cd>-texture-replacements/",
+      "Replace textures using HD versions from <cd>-texture-replacements/",
       NULL,
       "video",
       {
@@ -313,7 +313,7 @@ struct retro_core_option_v2_definition option_defs_us[] = {
       BEETLE_OPT(wireframe),
       "Wireframe Mode (Debug)",
       NULL,
-      "When enabled, renders 3D models in outline form without textures or shading. Only supported by the OpenGL hardware renderer. Note: This is for debugging purposes, and should normally be disabled.",
+      "Render 3D models in outline form without textures or shading. Only supported by the OpenGL hardware renderer. Note: This is for debugging purposes, and should normally be disabled.",
       NULL,
       "video",
       {
@@ -328,7 +328,7 @@ struct retro_core_option_v2_definition option_defs_us[] = {
       BEETLE_OPT(frame_duping),
       "Frame Duping (Speedup)",
       NULL,
-      "When enabled and supported by the libretro frontend, provides a small performance increase by directing the frontend to repeat the previous frame if the core has nothing new to display.",
+      "When enabled and supported by the libretro frontend, this provides a small performance increase by directing the frontend to repeat the previous frame if the core has nothing new to display.",
       NULL,
       "video",
       {
@@ -357,7 +357,7 @@ struct retro_core_option_v2_definition option_defs_us[] = {
       BEETLE_OPT(display_vram),
       "Display Full VRAM (Debug)",
       NULL,
-      "When enabled, visualizes the entire emulated console's VRAM. Only supported by the OpenGL and Vulkan hardware renderers. Note: This is for debugging purposes, and should normally be disabled.",
+      "Visualize the entire emulated console's VRAM. Only supported by the OpenGL and Vulkan hardware renderers. Note: This is for debugging purposes, and should normally be disabled.",
       NULL,
       "osd",
       {
@@ -372,7 +372,7 @@ struct retro_core_option_v2_definition option_defs_us[] = {
       BEETLE_OPT(analog_calibration),
       "Analog Self-Calibration",
       NULL,
-      "When the input device is set to DualShock, Analog Controller, Analog Joystick, or neGcon, this option allows dynamic calibration of analog inputs. Maximum registered input values are monitored in real time and used to scale analog coordinates passed to the emulator. This should be used for games such as Mega Man Legends 2 that expect larger values than what modern controllers provide. For best results, analog sticks should be rotated at full extent to tune the calibration algorithm each time content is loaded.",
+      "When the input device is set to DualShock, Analog Controller, Analog Joystick, or neGcon, this option enables dynamic calibration of analog inputs. Maximum registered input values are monitored in real time and used to scale analog coordinates passed to the emulator. This should be used for games such as Mega Man Legends 2 that expect larger values than what modern controllers provide. For best results, analog sticks should be rotated at full extent to tune the calibration algorithm each time content is loaded.",
       NULL,
       "input",
       {
@@ -386,7 +386,7 @@ struct retro_core_option_v2_definition option_defs_us[] = {
       BEETLE_OPT(analog_toggle),
       "Enable DualShock Analog Mode Toggle",
       NULL,
-      "When the input device type is DualShock, sets whether or not the emulated DualShock can be toggled between DIGITAL and ANALOG mode like original hardware. When this option is disabled, the DualShock is locked to ANALOG mode and when enabled, the DualShock can be toggled between DIGITAL and ANALOG mode by pressing and holding START+SELECT+L1+L2+R1+R2.",
+      "When the input device type is DualShock, this option allows the emulated DualShock to be toggled between DIGITAL and ANALOG mode like original hardware. When disabled, the DualShock is locked to ANALOG mode and when enabled, the DualShock can be toggled between DIGITAL and ANALOG mode by pressing and holding START+SELECT+L1+L2+R1+R2.",
       NULL,
       "input",
       {
@@ -400,7 +400,7 @@ struct retro_core_option_v2_definition option_defs_us[] = {
       BEETLE_OPT(enable_multitap_port1),
       "Port 1: Multitap Enable",
       NULL,
-      "Enables/Disables multitap functionality on port 1.",
+      "Enable multitap functionality on port 1.",
       NULL,
       "input",
       {
@@ -414,7 +414,7 @@ struct retro_core_option_v2_definition option_defs_us[] = {
       BEETLE_OPT(enable_multitap_port2),
       "Port 2: Multitap Enable",
       NULL,
-      "Enables/Disables multitap functionality on port 2.",
+      "Enable multitap functionality on port 2.",
       NULL,
       "input",
       {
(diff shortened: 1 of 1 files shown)


libretro_core_options.h
```diff
@@ -175,16 +175,16 @@ struct retro_core_option_v2_definition option_defs_us[] = {
       BEETLE_OPT(filter_exclude_sprite),
       "Exclude Sprites from Filtering",
       NULL,
-      "Do not apply texture filtering to sprites. Prevent seams in various games with 2D sprite-rendered backgrounds. Use together with Adaptive Smoothing or another post-processing filter for best effect.",
+      "Do not apply texture filtering to sprites. Prevents seams in various games with 2D sprite-rendered backgrounds. Use together with Adaptive Smoothing or another post-processing filter for best effect.",
```
IlDucci, line 178: Possibly the "various" word is unneeded. It should be obvious that the affected games are those with 2D sprite-rendered backgrounds, whether it's all of those games or not. Or maybe it could be changed to "certain" or "specific".

libretro_core_options.h
```diff
@@ -253,7 +253,7 @@ struct retro_core_option_v2_definition option_defs_us[] = {
       BEETLE_OPT(mdec_yuv),
       "MDEC YUV Chroma Filter",
       NULL,
-      "Improves the quality of FMV playback by reducing 'macroblocking' artifacts (squares/jagged edges). Only supported by the Vulkan renderer.",
+      "Improve the quality of FMV playback by reducing 'macroblocking' artifacts (squares/jagged edges). Only supported by the Vulkan renderer.",
```
IlDucci, line 256: I don't think the word "macroblocking" needs quotes. Maybe it was quoted to make it more visible for non-expert users, but isn't there another way to do that?

libretro_core_options.h
```diff
@@ -124,7 +124,7 @@ struct retro_core_option_v2_definition option_defs_us[] = {
       BEETLE_OPT(dither_mode),
       "Dithering Pattern",
       NULL,
-      "Set dithering pattern configuration. '1x (Native)' emulates native low resolution dithering used by original hardware to smooth out color banding artifacts visible at native color depth. 'Internal Resolution' scales dithering granularity to the configured internal resolution for cleaner results. Recommended to be disabled when running at 32 bpp color depth. Note: On Vulkan, enabling this will force downsampling to native color depth, while disabling will automatically enable output at higher color depth.",
+      "Choose dithering pattern configuration. '1x (Native)' emulates native low resolution dithering used by original hardware to smooth out color banding artifacts visible at native color depth. 'Internal Resolution' scales dithering granularity to the configured internal resolution for cleaner results. Recommended to be disabled when running at 32 bpp color depth. Note: On Vulkan, enabling this will force downsampling to native color depth, while disabling will automatically enable output at higher color depth.",
```
IlDucci, line 127: This is an overall warning of a probable lack of articles in some of the phrases. Examples in here: "emulates the native low resolution dithering"...

libretro_core_options.h
```diff
@@ -156,7 +156,7 @@ struct retro_core_option_v2_definition option_defs_us[] = {
       BEETLE_OPT(filter),
       "Texture Filtering",
       NULL,
-      "Select texture filtering method. 'Nearest' emulates original hardware. 'Bilinear' and '3-Point' are smoothing filters, which reduce pixelation via blurring. 'SABR', 'xBR', and 'JINC2' are upscaling filters that may improve texture fidelity/sharpness at the expense of increased performance requirements. Only supported by the hardware renderers.",
+      "Choose texture filtering method. 'Nearest' emulates original hardware. 'Bilinear' and '3-Point' are smoothing filters, which reduce pixelation via blurring. 'SABR', 'xBR', and 'JINC2' are upscaling filters that may improve texture fidelity/sharpness at the expense of increased performance requirements. Only supported by the hardware renderers.",
```
IlDucci, line 159: "'Nearest' emulates the original hardware."

libretro_core_options.h
```diff
@@ -372,7 +372,7 @@ struct retro_core_option_v2_definition option_defs_us[] = {
       BEETLE_OPT(analog_calibration),
       "Analog Self-Calibration",
       NULL,
-      "When the input device is set to DualShock, Analog Controller, Analog Joystick, or neGcon, this option allows dynamic calibration of analog inputs. Maximum registered input values are monitored in real time and used to scale analog coordinates passed to the emulator. This should be used for games such as Mega Man Legends 2 that expect larger values than what modern controllers provide. For best results, analog sticks should be rotated at full extent to tune the calibration algorithm each time content is loaded.",
+      "When the input device is set to DualShock, Analog Controller, Analog Joystick, or neGcon, this option enables dynamic calibration of analog inputs. Maximum registered input values are monitored in real time and used to scale analog coordinates passed to the emulator. This should be used for games such as Mega Man Legends 2 that expect larger values than what modern controllers provide. For best results, analog sticks should be rotated at full extent to tune the calibration algorithm each time content is loaded.",
```
IlDucci, line 375: "to scale the analog coordinates passed to"...? Possible lack of a comma after "Mega Man Legends 2"? Maybe to make things a bit more clearer, the "each time (a) content is loaded" should be taken to between "at full extent" and "to tune the calibration algorithm".

libretro_core_options.h
```diff
@@ -386,7 +386,7 @@ struct retro_core_option_v2_definition option_defs_us[] = {
       BEETLE_OPT(analog_toggle),
       "Enable DualShock Analog Mode Toggle",
       NULL,
-      "When the input device type is DualShock, sets whether or not the emulated DualShock can be toggled between DIGITAL and ANALOG mode like original hardware. When this option is disabled, the DualShock is locked to ANALOG mode and when enabled, the DualShock can be toggled between DIGITAL and ANALOG mode by pressing and holding START+SELECT+L1+L2+R1+R2.",
+      "When the input device type is DualShock, this option allows the emulated DualShock to be toggled between DIGITAL and ANALOG mode like original hardware. When disabled, the DualShock is locked to ANALOG mode and when enabled, the DualShock can be toggled between DIGITAL and ANALOG mode by pressing and holding START+SELECT+L1+L2+R1+R2.",
```
IlDucci, line 389: Possibly lacks a comma after "and ANALOG mode".

libretro_core_options.h
```diff
@@ -428,7 +428,7 @@ struct retro_core_option_v2_definition option_defs_us[] = {
       BEETLE_OPT(gun_input_mode),
       "Gun Input Mode",
       NULL,
-      "Specify whether to use a mouse-controlled 'Light Gun' or a 'Touchscreen' input when device type is set to 'Guncon/G-Con 45' or 'Justifier'.",
+      "Choose whether to use a mouse-controlled 'Light Gun' or a 'Touchscreen' input when device type is set to 'Guncon/G-Con 45' or 'Justifier'.",
```
IlDucci, line 431: "when the device type is set"...? "Touch Screen"? Also, one thing to keep in mind for the future: those controller types are unlocalizable at this moment.

libretro_core_options.h
```diff
@@ -443,13 +443,13 @@ struct retro_core_option_v2_definition option_defs_us[] = {
       BEETLE_OPT(gun_cursor),
       "Gun Cursor",
       NULL,
-      "Select the gun cursor to be displayed on screen while using the 'Guncon/G-Con 45' and 'Justifier' input device types. When disabled, crosshairs are always hidden.",
+      "Choose the gun cursor to be displayed on screen while using the 'Guncon/G-Con 45' and 'Justifier' input device types. When disabled, crosshairs are always hidden.",
```
IlDucci, line 446: I think some people would not identify a cursor as a "crosshair" or "sight", as commonly written in other emulators. What do you think?

libretro_core_options.h
```diff
@@ -458,7 +458,7 @@ struct retro_core_option_v2_definition option_defs_us[] = {
       BEETLE_OPT(crosshair_color_p1),
       "Port 1: Gun Crosshair Color",
       NULL,
-      "Change the light gun crosshair color for port 1.",
+      "Choose the light gun crosshair color for port 1.",
```
IlDucci, line 461: Unification issue: cursor or crosshair. Affects the setting label as well.

libretro_core_options.h
```diff
@@ -480,12 +480,12 @@ struct retro_core_option_v2_definition option_defs_us[] = {
       BEETLE_OPT(crosshair_color_p2),
       "Port 2: Gun Crosshair Color",
       NULL,
-      "Change the light gun crosshair color for port 2.",
+      "Choose the light gun crosshair color for port 2.",
```
IlDucci, line 483: Unification issue: cursor or crosshair. Affects the setting label as well.

libretro_core_options.h
```diff
@@ -602,7 +602,7 @@ struct retro_core_option_v2_definition option_defs_us[] = {
       BEETLE_OPT(enable_memcard1),
       "Enable Memory Card 1 (Restart Required)",
       NULL,
-      "Select whether to emulate a second memory card in slot 1. When disabled, games can only access the memory card in slot 0. Note: Some games require this option to be disabled for correct operation (e.g. Codename Tenka). Note: Memory Card 1 uses the Mednafen (.mcr) save format.",
+      "Emulate a second memory card in slot 1. When disabled, games can only access the memory card in slot 0. Note: Some games require this option to be disabled for correct operation (e.g. Codename Tenka). Note: Memory Card 1 uses the Mednafen (.mcr) save format.",
```
IlDucci, line 605: Lowercase the second "Memory Card" to unify all cases. Seems that Sony didn't trademark the words "Memory Card", so they don't require mandatory uppercasing.

libretro_core_options.h
```diff
@@ -630,7 +630,7 @@ struct retro_core_option_v2_definition option_defs_us[] = {
       BEETLE_OPT(memcard_left_index),
       "Memory Card Left Index",
       NULL,
-      "Changes the memory card currently loaded in the left slot. This option will only work if Memory Card 0 method is set to Mednafen. The default card is index 0.",
+      "Change the memory card currently loaded in the left slot. This option will only work if Memory Card 0 method is set to Mednafen. The default card is index 0.",
```
IlDucci, line 633: Lowercase the second "Memory Card" to unify all cases. Seems that Sony didn't trademark the words "Memory Card", so they don't require mandatory uppercasing. Also, "Mednafen" should be quoted to indicate that it is an option.

libretro_core_options.h
```diff
@@ -706,12 +706,12 @@ struct retro_core_option_v2_definition option_defs_us[] = {
       BEETLE_OPT(memcard_right_index),
       "Memory Card Right Index",
       NULL,
-      "Changes the memory card currently loaded in the right slot. This option will only work if Memory Card 1 is enabled. The default card is index 1.",
+      "Change the memory card currently loaded in the right slot. This option will only work if Memory Card 1 is enabled. The default card is index 1.",
```
IlDucci, line 709: Lowercase the second "Memory Card" to unify all cases. Seems that Sony didn't trademark the words "Memory Card", so they don't require mandatory uppercasing.

libretro_core_options.h
```diff
@@ -834,7 +834,7 @@ struct retro_core_option_v2_definition option_defs_us[] = {
       BEETLE_OPT(pgxp_vertex),
       "PGXP Vertex Cache",
       NULL,
-      "Allows PGXP-enhanced vertex positions to be cached for re-use across polygon draws. Can potentially improve object alignment and reduce visible seams when rendering textures, but false positives when querying the cache may produce graphical glitches. It is currently recommended to leave this option disabled. This option is applied only when PGXP Operation Mode is enabled. Only supported by the hardware renderers.",
+      "Cache PGXP-enhanced vertex positions for re-use across polygon draws. Can potentially improve object alignment and reduce visible seams when rendering textures, but false positives when querying the cache may produce graphical glitches. It is currently recommended to leave this option disabled. This option is applied only when PGXP Operation Mode is enabled. Only supported by the hardware renderers.",
```
IlDucci, line 837: "by the hardware renderers" needs the article?

libretro_core_options.h
```diff
@@ -848,7 +848,7 @@ struct retro_core_option_v2_definition option_defs_us[] = {
       BEETLE_OPT(pgxp_texture),
       "PGXP Perspective Correct Texturing",
       NULL,
-      "When enabled, replaces native PSX affine texture mapping with perspective correct texture mapping. Eliminates position-dependent distortion and warping of textures, resulting in properly aligned textures. This option is applied only when PGXP Operation Mode is enabled. Only supported by the hardware renderers.",
+      "Replace native PSX affine texture mapping with perspective correct texture mapping. Eliminates position-dependent distortion and warping of textures, resulting in properly aligned textures. This option is applied only when PGXP Operation Mode is enabled. Only supported by the hardware renderers.",
```
IlDucci, line 851: "by the hardware renderers" needs the article?

libretro_core_options.h
```diff
@@ -892,12 +892,12 @@ struct retro_core_option_v2_definition option_defs_us[] = {
       BEETLE_OPT(widescreen_hack_aspect_ratio),
       "Widescreen Mode Hack Aspect Ratio",
       NULL,
-      "The aspect ratio that's used by the Widescreen Mode Hack.",
+      "Choose the aspect ratio to be used by the Widescreen Mode Hack.",
```
IlDucci, line 895: Maybe "the aspect ratio to apply/force with the Widescreen Mode Hack"?

libretro_core_options.h
```diff
@@ -911,7 +911,7 @@ struct retro_core_option_v2_definition option_defs_us[] = {
       BEETLE_OPT(cpu_freq_scale),
       "CPU Frequency Scaling (Overclock)",
       NULL,
-      "Enable overclocking (or underclocking) of the emulated PSX CPU. Overclocking can eliminate slowdown and improve frame rates in certain games at the expense of increased performance requirements. Note that some games have an internal frame rate limiter and may not benefit from overclocking. May cause certain effects to animate faster than intended in some titles when overclocked.",
+      "Overclock (or underclock) the emulated PSX CPU. Overclocking can eliminate slowdown and improve frame rates in certain games at the expense of increased performance requirements. Note that some games have an internal frame rate limiter and may not benefit from overclocking. May cause certain effects to animate faster than intended in some titles when overclocked.",
```
IlDucci, line 914: "eliminate slowdowns", "at the expense" is adequate?

libretro_core_options.h
```diff
@@ -1008,7 +1008,7 @@ struct retro_core_option_v2_definition option_defs_us[] = {
       BEETLE_OPT(skip_bios),
       "Skip BIOS",
       NULL,
-      "Skips the PlayStation BIOS boot animation normally displayed when loading content. Note: Enabling this causes compatibility issues with a number of games (PAL copy protected games, Saga Frontier, etc.).",
+      "Skip the PlayStation BIOS boot animation normally displayed when loading content. Note: Enabling this causes compatibility issues with a number of games (PAL copy protected games, Saga Frontier, etc.).",
```
IlDucci, line 1011: "when loading content" lacks the "a" article. Would that be enough to make it better in English? I kinda have the feeling that something is still off there.

libretro_core_options.h
```diff
@@ -1038,7 +1038,7 @@ struct retro_core_option_v2_definition option_defs_us[] = {
...
       NULL,
-      "Choose video renderer. The software renderer is the most accurate but has steep performance requirements when running at increased internal GPU resolutions. The hardware renderers, while less accurate, improve performance over the software renderer at increased internal resolutions and enable various graphical enhancements. 'Hardware (Auto)' automatically selects the Vulkan or OpenGL renderer, depending upon the current libretro frontend video driver. If the provided video driver is not Vulkan or OpenGL 3.3-compatible, then the core will fall back to the software renderer.",
+      "The software renderer is the most accurate but has steep performance requirements when running at increased internal GPU resolutions. The hardware renderers, while less accurate, improve performance over the software renderer at increased internal resolutions and enable various graphical enhancements. 'Hardware (Auto)' automatically selects the Vulkan or OpenGL renderer, depending upon the current libretro frontend video driver. If the provided video driver is not Vulkan or OpenGL 3.3-compatible, then the core will fall back to the software renderer.",
```
IlDucci, line 1041: Might lack a comma after "is the most accurate".

libretro_core_options.h
```diff
@@ -1125,11 +1125,11 @@ struct retro_core_option_v2_definition option_defs_us[] = {
       BEETLE_OPT(core_timing_fps),
       "Core-Reported FPS Timing",
       NULL,
-      "Sets FPS timing that the core will report to the frontend. Automatic toggling will allow the core to switch between reporting progressive and interlaced rates, but may cause frontend video/audio driver reinits.",
+      "Choose the FPS timing that the core will report to the frontend. Automatic Toggling will allow the core to switch between reporting progressive and interlaced rates, but may cause frontend video/audio driver re-inits.",
```
IlDucci, line 1128: "re-inits" might be confusing, perhaps?

libretro_core_options.h
```diff
@@ -1238,7 +1238,7 @@ struct retro_core_option_v2_definition option_defs_us[] = {
       BEETLE_OPT(image_offset_cycles),
       "Horizontal Image Offset (GPU Cycles)",
       NULL,
-      "Specify number of GPU cycles to offset image by. Positive values move image to the right, negative values move image to the left. Only supported by the hardware renderers.",
+      "Choose number of GPU cycles to offset image by. Positive values move image to the right, negative values move image to the left. Only supported by the hardware renderers.",
```
IlDucci, line 1241: "move the image", in both cases?

libretro_core_options.h
```diff
@@ -1350,7 +1350,7 @@ struct retro_core_option_v2_definition option_defs_us[] = {
       BEETLE_OPT(aspect_ratio),
       "Core Aspect Ratio",
       NULL,
-      "Set core provided aspect ratio. This setting is ignored when the Widescreen Mode Hack or Display Full VRAM options are enabled.",
+      "Choose core provided aspect ratio. This setting is ignored when the Widescreen Mode Hack or Display Full VRAM options are enabled.",
```
IlDucci, line 1353: The options could need quotation marks to follow existing customs.

libretro_core_options.h
```diff
@@ -1418,7 +1418,7 @@ struct retro_core_option_v2_definition option_defs_us[] = {
       BEETLE_OPT(last_scanline),
       "Last Scan Line - NTSC",
       NULL,
-      "Select the last displayed scan line when running NTSC content. Setting a value less than 239 will reduce the height of output images by cropping pixels from the bottommost edge. May be used to counteract letterboxing.",
+      "Choose the last displayed scan line when running NTSC content. Values less than 239 will reduce the height of output images by cropping pixels from the bottommost edge. May be used to counteract letterboxing.",
```
IlDucci, line 1421: Unification issue: "Values below 239"...

libretro_core_options.h
```diff
@@ -1584,7 +1584,7 @@ struct retro_core_option_v2_definition option_defs_us[] = {
       BEETLE_OPT(cd_access_method),
       "CD Access Method (Restart Required)",
       NULL,
-      "Select method used to read data from content disk images. 'Synchronous' mimics original hardware. 'Asynchronous' can reduce stuttering on devices with slow storage. 'Pre-Cache' loads the entire disk image into memory when launching content, which may improve in-game loading times at the cost of an initial delay at startup. 'Pre-Cache' may cause issues on systems with low RAM, and will fall back to 'Synchronous' for physical media",
+      "Choose method used to read data from content disk images. 'Synchronous' mimics original hardware. 'Asynchronous' can reduce stuttering on devices with slow storage. 'Pre-Cache' loads the entire disk image into memory when launching content, which may improve in-game loading times at the cost of an initial delay at startup. 'Pre-Cache' may cause issues on systems with low RAM, and will fall back to 'Synchronous' for physical media.",
```
IlDucci, line 1587: "Choose the method", "mimics the original hardware", "slow storage" is self-explanatory enough? "when launching a content".

libretro_core_options.h
```diff
@@ -1600,7 +1600,7 @@ struct retro_core_option_v2_definition option_defs_us[] = {
       BEETLE_OPT(cd_fastload),
       "CD Loading Speed",
       NULL,
-      "Select disk access speed multiplier. Setting this higher than '2x (Native)' can greatly reduce in-game loading times, but may introduce timing errors. Some games may not function properly if this option is increased above a certain value.",
+      "Choose disk access speed multiplier. Values higher than '2x (Native)' can greatly reduce in-game loading times, but may introduce timing errors. Some games may not function properly above a certain value.",
```
IlDucci, line 1603: "properly above a certain value" is correct English?

IlDucci (commented):
Review done. You will see a lot of questions rather than full mentions, but that's what happens when I'm not a good enough English speaker. Be specially wary of the lack of articles along the entire options file.

libretro_core_options.h
```diff
@@ -65,12 +65,12 @@ struct retro_core_option_v2_category option_cats_us[] = {
    {
       "input",
       "Input",
-      "Change lightgun, mouse and neGcon settings."
+      "Change light gun, mouse and neGcon settings."
```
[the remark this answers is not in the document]
DisasterMo: I just go by what seems to be the most common thing/the first thing a user should find on the matter; which is usually merriam-webster or, like in this case, wikipedia.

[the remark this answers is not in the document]
IlDucci: Yep, this definitely opens a can of worms.

libretro_core_options.h
```diff
@@ -90,7 +90,7 @@ struct retro_core_option_v2_definition option_defs_us[] = {
       BEETLE_OPT(internal_resolution),
       "Internal GPU Resolution",
       NULL,
-      "Set internal resolution multiplier. Resolutions higher than '1x (Native)' improve fidelity of 3D models at the expense of increased performance requirements. 2D elements are generally unaffected by this setting.",
+      "Choose internal resolution multiplier. Resolutions higher than '1x (Native)' improve fidelity of 3D models at the expense of increased performance requirements. 2D elements are generally unaffected by this setting.",
```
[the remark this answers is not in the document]
DisasterMo: Articles are technically not required in those cases. 
But let's say, we add one, which would we even use here, "the" or "a"? And why?

libretro_core_options.h
```diff
@@ -108,7 +108,7 @@ struct retro_core_option_v2_definition option_defs_us[] = {
       BEETLE_OPT(depth),
       "Internal Color Depth",
       NULL,
-      "Set internal color depth. Higher color depth can reduce color banding effects without the use of dithering. 16 bpp emulates original hardware but may have visible banding if dithering is not enabled. 'Dithering Pattern' is recommended to be disabled when this option is set to 32 bpp.",
+      "Choose internal color depth. Higher color depth can reduce color banding effects without the use of dithering. 16 bpp emulates original hardware but may have visible banding if dithering is not enabled. 'Dithering Pattern' is recommended to be disabled when this option is set to 32 bpp.",
```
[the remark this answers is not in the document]
DisasterMo: 1.: Yeah, can't believe I missed that.
2.: It's hard to explain why, but an article is not required here.
3.: Yes, this should probably be something like "but may cause visible color banding".
4.: Any suggestion on what you'd like to see instead?

libretro_core_options.h
```diff
@@ -140,7 +140,7 @@ struct retro_core_option_v2_definition option_defs_us[] = {
       BEETLE_OPT(scaled_uv_offset),
       "Texture UV Offset",
       NULL,
-      "Sample textures for 3D polygons at an offset for higher than 1x internal resolution. Reduce texture seams but may cause unintended graphical glitches.",
+      "Sample textures for 3D polygons at an offset for higher than 1x internal resolution. Reduces texture seams but may cause unintended graphical glitches.",
```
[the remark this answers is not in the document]
DisasterMo: In this case, wouldn't it be better to omit "1x" altogether?

libretro_core_options.h
DisasterMo: Yeah, you're right, that word is unnecessary fluff.

libretro_core_options.h
DisasterMo: Good catch!

libretro_core_options.h
DisasterMo: 1.: yes
2.: also, yes
3.: what do you mean? "on devices with slow storage" -> "on slow storage" ?
4.: absolutely not, this is the general case, no article required.

libretro_core_options.h
DisasterMo: It is. But I agree that this sounds strange. Do you have a better suggestion?
Maybe "Some games may not function properly if this value is too high."?

libretro_core_options.h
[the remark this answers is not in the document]
IlDucci: Okay, correction on my end: this opened up a can of worms... but not in English. The English texts used the proper "light gun", with two words, expression. My bad.

libretro_core_options.h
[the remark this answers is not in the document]
IlDucci: I think it would be a "the" in this case, as an option has to be chosen, even 1x, which is the default setting that does no changes from the original resolution. But this might be a case of me thinking a bit in Spanish, rather than in English. If it's not required, you can discard this comment.

libretro_core_options.h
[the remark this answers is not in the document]
IlDucci: For the fourth comment, I think the phrase is too long, without pauses or anything that gives it an easier structure to read. Not sure how to explain this one.

libretro_core_options.h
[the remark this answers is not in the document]
IlDucci: That works as well.

libretro_core_options.h
IlDucci: For the third situation, perhaps I don't have a use case with slow storage (I do have a PSVita, but I never use it with RA). I think this is another case of me thinking in Spanish, as what I am missing is an extra word after "storage", not that are superfluous words.

libretro_core_options.h
IlDucci: That sounds better for my eyes.

libretro_core_options.h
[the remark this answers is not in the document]
DisasterMo: Oh, yeah, the German term is 'Lightgun' in one word as well. That's not really a problem, though. Marking this as resolved.

libretro_core_options.h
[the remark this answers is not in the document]
DisasterMo: Okay, I'll try to think of something.

libretro_core_options.h
DisasterMo: Oh, something like "on devices with a slow storage medium"? But I feel like storage is enough on its own in this case (for English).

libretro_core_options.h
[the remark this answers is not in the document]
DisasterMo: Okay, I'll go over them again and try to add articles where it fits. Thank you for your input.

libretro_core_options.h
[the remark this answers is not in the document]
DisasterMo: Also: You are likely right with 2., it really needs an article.

libretro_core_options.h
DisasterMo: Who cares, the quotes don't belong either way.

libretro_core_options.h
DisasterMo: Both one word and two words variants of Touchscreen are acceptable (source: Merriam-Webster).
So, which one should we pick?

As for the controller types: where are those strings? RA or somewhere in this code?

libretro_core_options.h
DisasterMo: Let's stick to crosshair then? Seems like the more common term.

libretro_core_options.h
DisasterMo: Probably not. But it's also a good distinction "supported by the hardware renderers we provide, but not by the software renderers". It works either way. Might as well keep it as it is, right?

libretro_core_options.h
DisasterMo: "Choose the aspect ratio for the 'Widescreen Mode Hack'."?

libretro_core_options.h
DisasterMo: > "at the expense" is adequate?

What else do you suggest? "at the cost"?

libretro_core_options.h
IlDucci: Oh, okay then with Touchscreen. Seems that the commonly used word in RA is also Touchscreen (with the exception of the 3DS, but that has a motive), so that issue can be ignored.

As for the types... I don't know. Those typically appear when going to the Controls section in the Quick Menu or the main Input menu, so... That could lead to something bigger.

libretro_core_options.h
IlDucci: Okay then.

libretro_core_options.h
DisasterMo: > "when loading content" lacks the "a" article.

Once again, this is the general case. Let me explain. The full sentence is actually:
"Skip the PlayStation BIOS boot animation normally displayed when loading _any_ content"

Any can be omitted in some cases, such as above, but it cannot be replaced by an article.

libretro_core_options.h
IlDucci: Okay then, to be ignored then.

libretro_core_options.h
IlDucci: Much better and simpler.

libretro_core_options.h
DisasterMo: The best I can offer you rn is this:
Skip the PlayStation BIOS boot animation which is normally displayed when loading content.

libretro_core_options.h
DisasterMo: I'm not 100% sure, but maybe this'll work:
"[...] but may cause frontend video/audio drivers to restart."

PR #850, opened 2022-05-20
High resolution fmv video smoothing

MishaTrent:
I ask the developers very much to add support for smoothing fmv videos at high resolution to the kernel. Otherwise, they look terrible, you have to switch to adaptive smoothing. The YUV MDEC option does not work properly, in theory it should smooth.


Ryunam:
I think you probably wanted to open an issue, rather than send a pull request with more than 300 old commits…!

ds22x:
Why did you open up an issue as a pull request?

LibretroAdmin:
Closing this PR as invalid. Please resubmit if a proper PR can be devised.

PR #853, opened 2022-06-06
Check BIOSes by SHA1 and filesize, not filename.

mudlord:
Pretty simple. Removes all the headache of finding files by filename, and instead leaves it to SHA1 instead. Also checks by filesize to remove speed probs.


ZachCook:
Looks useful, should squash those commits though.

mudlord:
done, check the other req.

PR #856, opened 2022-07-07
GLES 3.0 Support on HW Renderer

BinBashBanana:
## Overview:
The OpenGL renderer was lacking GLES 3.0 support, so I did my best to add it.

Colors are not 100% perfect on GLES3 due to it not including the 1555 ABGR color format. GLES3 uses 5551 RGBA instead.
- example: loading screens on crash bandicoot are the wrong color

## List of changes:
- updated makefiles for:
    - android
    - ios
    - emscripten
    - rpi4

- tested on emscripten, GLES3 must be explicitly enabled for android and ios
    - emscripten: link with `-s TOTAL_MEMORY=536870912 -s FULL_ES3=1 -s MIN_WEBGL_VERSION=2 -s MAX_WEBGL_VERSION=2`
- fixed some things in glsm (libretro-common)
- use 5551 RGBA on GLES (shaders now convert from 1555 to 5551)


BinBashBanana:
Another thing I forgot, turning *off* dithering in GLES breaks it. This might be specific to emscripten, I'm not sure. WebGL complains about the vertex attribute not matching the type of the vertex shader input. This shouldn't really matter though. If all else fails, I can change the type from uint to float. (and the vertex attribute type)

LibretroAdmin:
Thanks very much for this!

basharast:
I was working on this implementation for UWP-ARM and mostly worked fine including shadows, colors..etc
I had to convert colors withing the code because I don't understand how shaders work
below attached my changes, 
I hope you already have it fixed but in case not, take look on my changes

PSX GLES: [rsx-.zip](https://github.com/libretro/beetle-psx-libretro/files/10778509/rsx-.zip)

PR #858, opened 2022-08-07
Change Dualshock Analog Mode Toggle Default

### LINKED ISSUE #738: [Feature Request] Allow user mapping of DualShock Analog Mode Toggle
It would be useful to allow custom configuration of `DualShock Analog Mode Toggle`, especially if users have an extra button to spare, or they wish to map to the middle button on their controller, retroarch menu defaults aside.

Also, the default mapping of `Start+Select+R1+L1+R2+L2` is also the command for soft-reset in many games, some of which only require a single press to reset.

ichee:
Changes the default analog mode toggle combo buttons from `Start+Select+L1+L2+R1+R2` to `L1+R1`.

Prevents accidentally performing a soft reset in games that supported it.

Effectively fixes #738.

### DIFF
--- a/mednafen/psx/input/dualshock.cpp
+++ b/mednafen/psx/input/dualshock.cpp
@@ -174,7 +174,7 @@ void InputDevice_DualShock::CheckManualAnaModeChange(void)
 
       if(amct_enabled)
       {
-         if(buttons[0] == 0x09 && buttons[1] == 0x0f)
+         if(buttons[0] == 0x00 && buttons[1] == 0x0c)
          {
             if(combo_anatoggle_counter == -1)
                combo_anatoggle_counter = 0;


dankcushions:
surely this would be triggered accidentally too often? most games probably use these buttons in gameplay

ichee:
No games I'm aware require both pressing both buttons at the same time, nor do any require holding them for an extended period, beyond a soft reset which is the main point of this PR.

The stick buttons would be more ideal, but it's been reported that not all controllers or handhelds that are supported have stick buttons, so this is the best compromise I can think of.

I can increase the hold time to toggle analog to 2 or 3 seconds to minimize any chance of accidentally triggering if needed.

hizzlekizzle:
This would definitely be easy to trigger accidentally in Armored Core during a regular fight. What about adding start or select to it? that way someone would have to move their hand off of the controls altogether to trigger it.

ichee:
I can do that.

Would you prefer also increasing the hold time, or leaving it as is?

ichee:
Anything else that needs resolving here for merging to be considered?

bslenul:
Sounds good, I always thought the default combo was kinda annoying.

However, please update the combo from the option description: https://github.com/libretro/beetle-psx-libretro/blob/0763daa1dae146d1cfd118265e1c6cdede0e8348/libretro_core_options.h#L389

ichee:
Done.

And thanks for pointing that out. I had forgotten about the description.

ichee:
Closing in favor of #859.

PR #859, opened 2022-08-20
Add multiple combos for the analog toggle

### LINKED ISSUE #699: [Feature Request] Alternate analog toggle chord
Right now, the only chord available for analog toggle is L1+L2+R1+R2+Select+Start. This is also the chord used internally by a significant number of PS1 games to trigger a soft reset. This makes it completely non-viable for a number of games. I'd like to propose a second option under "Enable Dualshock Analog Mode Toggle" to select the preferred button chord. Good options would include L1+L2+L3, R1+R2+R3, or even L1+L2+R3 and R1+R2+L3 for people with less manual dexterity. The retroarch menu toggle list has some other good options as well.

Personally, I'd go with R1+R2+R3 for my own use. It's easy to enter and shouldn't conflict with any game or any currently available chord.

There's also the prospect of an option to enable analog to digital on the left stick automatically when analog mode is toggled off in this fashion, but that may be a topic for another issue.

bslenul:
As an alternative to #858 I'm proposing a new core option that lets you chose a combo from a predefined list, instead of hardcoding a single one:

image

image

I say that in pretty much every PR I make but I'm not a dev and there is maybe a better way to handle that, so feel free to edit the PR or to tell me what to change ;) But from my tests and @Sanaki's, it works perfectly fine!

Closes #699

### DIFF
--- a/libretro.cpp
+++ b/libretro.cpp
@@ -3139,6 +3139,8 @@ static bool shared_memorycards = false;
 static bool has_new_geometry = false;
 static bool has_new_timing = false;
 
+uint8_t analog_combo[2] = {0};
+
 extern void PSXDitherApply(bool);
 
 static void check_variables(bool startup)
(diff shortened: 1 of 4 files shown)


bslenul:
As Sanaki suggested to me on Discord, I changed it so it's now a separate option, to not break existing core options files.

bslenul:
Rebased + very minor cleanups.

PR has been pending for 6 months now, if there's anything wrong with it please let me know.

hizzlekizzle:
@LibretroAdmin looks safe to merge AFAICT

ichee:
Looks good.

I would still prefer if the default toggle was something other than `L1+L2+R1+R2+Start+Select`.

PR #860, opened 2022-09-11
fix #507 again

### LINKED ISSUE #507: [Beetle PSX HW] Android core is not actually HW but regular version
## Description

When you update the HW core via online updater on android, on either version of RA, the downloaded core is actually the normal version. It uses uses the same configuration and saves to the same directories as the normal version, even though the PC port uses other directories and is completely separate.

### Expected behavior

When loading the core, I expect it to say "Beetle PSX HW 0.99[whatever]".
Also, the config should be different than the normal version and saves should go to the "Beetle PSX HW" folders under saves and states.

### Actual behavior

The core appears identical to the normal, non HW version.

### Environment information

I'm using Samsung Galaxy Note 9, but I don't think it's relevant.

NOTE: This may not be relevant to this github. If it isn't, I would appreciate someone pointing me in the right direction. As far as I know, the RA buildbot doesn't have an address for such things.

BinBashBanana:
fixed the regression in android builds caused by [`3f525d7`](https://github.com/libretro/beetle-psx-libretro/commit/3f525d7d67b7c8f83ca0c951c279abc04decb17f)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -176,7 +176,7 @@ else ifneq (,$(findstring ios,$(platform)))
    ifeq ($(HAVE_OPENGL),1)
       GL_LIB := -framework OpenGLES
       GLES = 1
-      GLES3 = 0
+      GLES3 = 1
    endif
 
    CC = cc -arch $(iarch) -isysroot $(IOSSDK)
@@ -352,6 +352,7 @@ else ifeq ($(platform), rpi4_64)
    LDFLAGS += $(PTHREAD_FLAGS) -ldl -lrt
    HAVE_LIGHTREC = 1
    FLAGS += -DHAVE_SHM
+   GLES = 1
    GLES3 = 1
    GL_LIB := -lGLESv2
    HAVE_CDROM = 0
--- a/jni/Android.mk
+++ b/jni/Android.mk
@@ -2,10 +2,6 @@ LOCAL_PATH := $(call my-dir)
 
 CORE_DIR := $(LOCAL_PATH)/..
 
-# set these
-HAVE_HW = 0
-HAVE_OPENGLES3 = 0
-
 DEBUG                    := 0
 FRONTEND_SUPPORTS_RGB565 := 1
 NEED_CD                  := 1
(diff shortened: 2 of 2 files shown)


LibretroAdmin:
Hi, builds are broken now for Android ARMv7 and Android x86. Something about GLESv3 linkage. See below -

https://git.libretro.com/libretro/beetle-psx-libretro/-/pipelines/135422

BinBashBanana:
It looks like GLESv3 is only supported with android SDK 18 and later, see [here](https://stackoverflow.com/a/49142001). I have android SDK 33, and all 4 HW builds were successful on my computer. This could be fixed by making an exception for 32-bit android architectures, but would it be possible to upgrade your SDK minimum version on the [buildbot](https://git.libretro.com/libretro-infrastructure/ci-templates/-/blob/master/android-jni.yml) to at least 18? (It also turns out that SDK version 21 is always the minimum version for 64-bit builds, which explains why they were able to build, see the comment on the stack overflow answer)

LibretroAdmin:
We try to aim for as low a SDK level as possible for the sake of backwards compatibility with older Android versions.

You can proceed with the selective bumping of the API level for these two ABI targets

BinBashBanana:
Is it possible to do that within [.gitlab-ci.yml](https://github.com/libretro/beetle-psx-libretro/blob/9724fe15c36bf582c21cb6e36fc8610d4b1bdb79/.gitlab-ci.yml#L170)?

LibretroAdmin:
Don't know, you'd have to tell us, not the other way around.

BinBashBanana:
For now, I've just removed the OpenGL renderer from 32-bit android builds (armeabi-v7a and x86). I'll see if I can get a proper solution working for this though. PR: #861

LibretroAdmin:
OK, thanks. Hopefully something can be figured out, yeah. Maybe there'd be a way of doing some linking against a GLES shim and then the actual linking takes place on the frontend (RetroArch)?

PR #888, opened 2023-06-06
Run the content for a few frames before closing tray

### LINKED ISSUE #793: [Suggestion] Resume content automatically for a few frames when ejecting disc?
## Description

I think it would be great if we didn't have to manually resume content after ejecting a disc before inserting a new disc, I feel like it's not super intuitive for most users.

Maybe automatically resume the game (in the background) when ejecting disc? Like just for a few frames until the core "registers" that the disc was ejected? I have no idea how these cores handle that, but resuming is not required with PCSX ReARMed and Duck/SwanStation.

### Expected behavior

No resuming required:

* Go to `Quick Menu > Disc Control` and click "Eject".
* Switch disc index.
* Click "Insert".

### Actual behavior

Resuming is required (maybe not for every game? not sure):

* Go to `Quick Menu > Disc Control` and click "Eject".
* Resume.
* Go back to disc control menu and switch disc index.
* Click "Insert".

1. Start Metal Gear Solid from disc 2.
2. Select "New game".
3. Try switching disc without resuming, you'll get stuck at the "Insert disc 1" screen.

bslenul:
As explained in #793 currently when switching disc you have to resume content before actually closing the virtual tray or else it won't register that it was opened in the first place. This was frustrating and super confusing for new users, resuming doesn't feel intuitive at all and it's not needed with the other PS1 cores.

With this PR the content will run for a few frames before actually closing the tray, giving the core some time to register that it was opened before.

Closes #793

### DIFF
--- a/libretro.cpp
+++ b/libretro.cpp
@@ -85,6 +85,7 @@ static bool gui_inited = false;
 static bool gui_show = false;
 static char bios_path[4096];
 static bool firmware_found = false;
+static int close_tray_counter = 0;
 
 // Switchable memory cards
 static int memcard_left_index = 0;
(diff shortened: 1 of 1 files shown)


bslenul:
Btw, while testing stuff for this I noticed that switching discs is completely broken when runahead is enabled, saying just in case but it has nothing to do with this PR, it was already broken before :p I don't use runahead and I don't know much about it and how it works so unfortunately I have no idea on how to fix this.

PR #905, opened 2024-07-10
Implement neGcon controller profile with rumble

DarthMew:
This PR adds a separate neGcon controller profile with rumble support (a modified DualShock profile with neGcon identification and inputs), which was implemented in SwanStation through https://github.com/libretro/swanstation/commit/83776ae7d3b9221b2746f91af0c2d8d9b0994aa7, and is also available in its upstream counterpart and the MiSTer PSX core.
Main use case for this is that quite a handful of neGcon supported games also support sending rumble commands when in neGcon mode (mainly with third party steering wheels in mind).


DarthMew:
If anything else needs to be done to get this merged, please let me know.

PR #909, opened 2024-09-15
osx-arm64: disable lightrec

warmenhoven:
On osx arm64 the lightrec seems to be causing several memory corruption issues. Disabling the lightrec makes them all go away. Also the Mac seems to be performant enough not to strictly need it. I'm going to work on fixing and re-enabling the lightrec but until then this is a reasonable workaround.


warmenhoven:
I didn't even see the other pr until just now.

PR #912, opened 2024-09-25
Update libchdr + add zstd

### LINKED ISSUE #897: Update of CHD to support ZSTD
https://docs.mamedev.org/tools/chdman.html#createraw

Mame 0.262 introduced zstd as data compression. It would be good to update to the newest chd version to use the benefits from zstd over zlib.

sonninnos:
Closes #897


LibretroAdmin:
I've explained in a DM some followup steps that need to be taken:

* Replace all occurrences of 'inline' with 'INLINE'
* Fix all C89 build errors in libchdr files. To be more clear, put variable declarations at the top of a codeblock or function.

sonninnos:
Sorry for the delay, and hopefully I didn't miss anything.

negativeExponent:
hope the changes for the libchdr for libreto's use is universal, so that other core would take advantage of it as well in a more consistent matter. i have sent the same updates to the pc-engine line of cores and that has not recieve any feedback since or what to do about it. people have waited too long for the feature.

PR #945, opened 2026-03-17
Fix HW second instance runahead crashing and not rendering

### LINKED ISSUE #701: Current version of Beetle HW Crashes
## Description

(Using RetroArch)
When loading any content with the Beetle HW core, it crashes (and RA crashes too). This doesn't happen with the non-HW core. I am using the `vulkan` video driver.

If I switch to `gl` or `glcore`, the core doesn't crash, but it instead doesn't render anything (and I am able to use the RA quick menu to exit). If I switch to some other core (e.g. `d3d12`), it defaults to the software renderer and everything works fine.

### Expected behavior

I expect to load the content.

### Actual behavior

RA/Beetle HW core crash. RA crashes with an error of 
`Exception code=0xc0000005 flags=0x0 at 0x000000008052759E. Access violation - attempting to read data at address 0x0000000000000DB0`

I've attached the RA log, though it doesn't seem to print any errors.
[retroarch__2020_08_09__02_43_53.log](https://github.com/libretro/beetle-psx-libretro/files/5047119/retroarch__2020_08_09__02_43_53.log)

1. Load any content in RetroArch with Beetle HW core

### Bisect results

I was able to previously load content using Beetle HW as soon as July 3rd. Today is the first time I tried to load any content since then. I did update cores between then and now. I did not update RetroArch versions since then, but I tried updating it when I saw this error and results did not change.

### Version/Commit

Using the current latest version of Beetle PSX HW that is downloadable from RetroArch. This is version: 0.9.44.1 4cda2d4

### Core options

I use a set of different options, however, the same error occurs if I use default options. Here are those changed options regardless:
`
beetle_psx_hw_adaptive_smoothing = "enabled"
beetle_psx_hw_analog_calibration = "enabled"
beetle_psx_hw_dither_mode = "disabled"
`

### Environment information

- Frontend: RetroArch tried with both 1.8.8 (stable) and 1.9.0 (stable)
- OS: Windows 10 Pro (10.0.18363)
- Compiler: Using latest build from RA
- GPU: NVIDIA GeForce GTX 980 Ti / GeForce Game Ready Driver 451.67

ZachCook:
Stop vulkan renderer crashing with second instance runahead

If the context isn't initialized fall back to software, which also
fixes the second instance in runahead that never gets initialized

Fixes #701

### DIFF
--- a/rsx/rsx_intf.cpp
+++ b/rsx/rsx_intf.cpp
@@ -30,7 +30,7 @@
 extern bool fast_pal;
 extern unsigned int image_height;
 
-static enum rsx_renderer_type rsx_type          = RSX_SOFTWARE;
+enum rsx_renderer_type rsx_type          = RSX_SOFTWARE;
 
 static bool gl_initialized                      = false;
 static bool vk_initialized                      = false;
--- a/rsx/rsx_lib_gl.cpp
+++ b/rsx/rsx_lib_gl.cpp
@@ -72,6 +72,7 @@ static const GLushort indices[6] = {0, 1, 2, 1, 2, 3};
 #define VRAM_HEIGHT 512
 #define VRAM_PIXELS (VRAM_WIDTH_PIXELS * VRAM_HEIGHT)
 
+extern enum rsx_renderer_type rsx_type;
 extern retro_log_printf_t log_cb;
 
 /* How many vertices we buffer before forcing a draw. Since the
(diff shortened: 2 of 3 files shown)


LibretroAdmin:
Backported PR here so that it realigns with the current master
https://github.com/libretro/beetle-psx-libretro/commit/ce34f88c1dc8000c1a4b4092b3fc2f0298ff53b6
https://github.com/libretro/beetle-psx-libretro/commit/e30a2eba71680ae233d3a4ef4ba3cc8c5f835fde

PR #862, opened 2022-09-25
Lightning+Lightrec updates and other optimizations

ZachCook:
~~Split out cpu_lightrec.cpp from cpu.cpp instead of being mashed together.~~

Added hugepage, codebuffer, mmap at 0x0, and SP GP hit RAM optimizations and fallbacks, see commits for more details.

Use lightrec's GTE registers directly when not using beetle interpreter (while maintaining backwards compatibility with savestates).

Removed the Cycle Timing Check dynarec option.

Fixes #942

PR #957, opened 2026-05-14
Update Lightning and Lightrec subrepos, add some other optimizations

### LINKED ISSUE #942: Please update embbeded GNU lightning
That would help build this project on more exotic Debian architectures like loong64 or x32
Thanks

ZachCook:
Renamed lightning-lightrec-include folder to just be called include. 

Rewrote the mmap code while adding hugepage and codebuffer support.

Added SP GP hit RAM optimization, can generate faster code for lightrec, but can cause crashes in some titles, so is off by default, only use when needing every last bit of performance.

Rewrite GTE code to use lightrec's GTE registers directly when not using beetle interpreter (while maintaining backwards compatibility with savestates), most of the old variables become defines.

Removed the "Cycle Timing Check" dynarec option, lightrec's function for this was removed, could keep this if we have to, but as there is no reason to use it other than in-depth debugging, it doesn't need to be an option for users.

Fixes https://github.com/libretro/beetle-psx-libretro/issues/942  new platform support in updated Lightning.

See older version of this pull request: https://github.com/libretro/beetle-psx-libretro/pull/862


ZachCook:
Before merging I would like for someone to check for build and runtime errors on Windows and Mac, if so please show output of the mmap addresses with the info log level, which should look something like:

[libretro INFO] Lightrec map addresses: M=0x20000000, P=0x563c4a6887b0, R=0x3fc00000, H=0x3f800000
[libretro INFO] Lightrec codebuffer address: 0x20800000, size: 8MB (0x00800000)


I haven't seen any regression on the dozen or so games I tried, runahead, savestates, etc seem to work.

Be on the lookout for any missing or incorrect geometry as the GTE changes are quite substantial and could easily hide a few bugs since I only just updated it to build again against latest master branch, the code was written a while ago.

LibretroAdmin:
hi @ZachCook , the only rebase necessary here is because of the codebase finally using stdint everywhere. So that is the only change necessary for the conflicts to be resolved. Sorry about that.

ZachCook:
I ended up trying out Github's browser editor for conflict resolution for some reason, and added a few more u32 to uint32_t conversions in this branch's code, then ended up manually adding a missing conversion too later after it didn't build. I don't really like that local git shows slightly different diff and stats, but it does work, guess I can live with it if you can.

LibretroAdmin:
Checked out the branch, doesn't compile out of the box on Windows

libretro-common/vfs/vfs_implementation.c: In function 'retro_vfs_file_open_impl':
libretro-common/vfs/vfs_implementation.c:518:33: error: 'PROT_READ' undeclared (first use in this function); did you mean 'OF_READ'?
  518 |                stream->mapsize, PROT_READ,  MAP_SHARED, stream->fd, 0)) == MAP_FAILED)
      |                                 ^~~~~~~~~
      |                                 OF_READ
libretro-common/vfs/vfs_implementation.c:518:33: note: each undeclared identifier is reported only once for each function it appears in
libretro-common/vfs/vfs_implementation.c:518:45: error: 'MAP_SHARED' undeclared (first use in this function); did you mean 'LR_SHARED'?
  518 |                stream->mapsize, PROT_READ,  MAP_SHARED, stream->fd, 0)) == MAP_FAILED)
      |                                             ^~~~~~~~~~
      |                                             LR_SHARED
libretro-common/vfs/vfs_implementation.c:518:76: error: 'MAP_FAILED' undeclared (first use in this function); did you mean 'WAIT_FAILED'?
  518 |                stream->mapsize, PROT_READ,  MAP_SHARED, stream->fd, 0)) == MAP_FAILED)
      |                                                                            ^~~~~~~~~~
      |                                                                            WAIT_FAILED

ZachCook:
Looks like HAVE_MMAP for lightning (Makefile.common:   FLAGS += -DHAVE_MMAP) escaped its intended scope and messed up vfs, looks like it was a redundant define anyways? Does it build if you remove Makefile.common:   FLAGS += -DHAVE_MMAP

ZachCook:
@LibretroAdmin Looks like Mac needs another fix to build, maybe just ifndef define's to let those code path's fail like they should to fallback to the shm_open path, let me know if you need me create a patch for it.

